PR #17, opened 2015-02-27
handle projects that do not have package.json in root

conord33:
Compares the directory containing .git and the directory of containing husky to find the directory of the package.json file.  The relative path derived from this comparison is then added to the scripts in the .git/hooks files.  

This allows npm run --json to execute properly, instead of resulting in the following error:
```
npm ERR! npm run-script [<pkg>] <command>
npm ERR! not ok code 0
```

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -33,8 +33,15 @@ module.exports = {
       data += 'PATH="' + process.env.PATH + '"\n'
     }
 
+    // Get the path of the projects .git
+    var projectPath = dir.replace('.git/hooks', '')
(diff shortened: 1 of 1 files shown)


ppatriotis:
+1

typicode:
Thanks! :+1:

typicode:
Just released your feature in `v0.7.0`.
Did a little refactor to support Windows since PWD environment variable isn't available.

conord33:
nice good catch!

PR #21, opened 2015-06-16
Add post-checkout hook

spalger:
Added post-checkout hook.

I plan to write a little lib that checks for changes in package.json or bower.json files and offering to updated dependencies.


spalger:
Closing this for now.

The hook doesn't receive the positional args and doesn't seem to get called.

PR #23, opened 2015-07-21
Added post-checkout hook

mr-mig:
Added `post-checkout` hook.
We are using this stuff to run some build steps on every branch change.


typicode:
Thanks for the PR and using husky :)

I've just refactored it to support all hooks (including `post-checkout`). This way, there won't be anymore situations where a hook is missing.

PR #28, opened 2015-08-25
pass $1 to script if hook is commitmsg

jedverity:
There is definitely a better way to do this, passing any args along to the script to consider or ignore as necessary, but this should let users actually act on the commit message when using this hook.


typicode:
Good idea, thank you :)

typicode:
Just wanted to let you know I did a rollback.

I've extended your idea to other hooks, because I don't really want to have specific rules for some hooks.
But it introduced this bug then https://github.com/typicode/husky/issues/29

For the `commitmsg` hook, an alternative would be to directly read `.git/COMMIT_EDITMSG`.

Issue #29, opened 2015-09-04
v0.10.0: gulp is trying to use “origin” as a task


mistadikay:
Hi,

after updating to v0.10.0, I have a weird behavior in my prepush hook which is the following:

```json
{
  "scripts": {
    "prepush": "gulp --gulpfile gulpfile.js test:build"
  }
}
```

Where `test:build` is the name of gulp task. When I run it as `npm run prepush`, everything is fine. But when husky is running it on `git push origin master`, I have the following error:

```sh
[10:34:31] Using gulpfile ~/*PROJECT_DIRECTORY*/gulpfile.js
[10:34:31] Task never defined: origin
[10:34:31] Please check the documentation for proper gulpfile formatting

husky - pre-push hook failed (add --no-verify to bypass)
```

Which means that for some reason I have `origin` parameter leaked into npm script and it's trying to find gulp task named `origin`. Probably it's because of [this commit](https://github.com/typicode/husky/commit/c7ae4e070811451ecda3f586222ee04ad889dc33)?

typicode:
My bad, you're right. Since pre-commit takes no parameter, I just thought it would be the same with pre-push.

I'll fix that.

mistadikay:
@typicode `v0.10.1` works well, thanks!

PR #30, opened 2015-09-23
run npm scripts with arguments

deepsweet:
because for example in `prepare-commit-msg` hook `$1` means current message file, so we need to be able to change it.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -57,7 +57,7 @@ module.exports = {
       '[ -f package.json ] && cat package.json | grep -q \'"' + cmd + '"\\s*:\'',
       // package.json or script can't be found exit
       '[ $? -ne 0 ] && exit 0',
-      'npm run ' + cmd + ' --silent',
+      'npm run ' + cmd + ' --silent $@',
       'if [ $? -ne 0 ]; then',
       '  echo',
       '  echo "husky - ' + name + ' hook failed (add --no-verify to bypass)"',


deepsweet:
oh, I just saw https://github.com/typicode/husky/commit/35386f0de9c46f99e679e47d63c1dfd994d2954f

why you removed it?

typicode:
Hi,

Sorry for the rollback, I replied in https://github.com/typicode/husky/pull/28 to let you know about it.
It's mostly because of https://github.com/typicode/husky/issues/29.

Actually, passing git parameters to scripts can introduce unexpected behaviors.
I'll take the example in https://github.com/typicode/husky/issues/29:

```json
"prepush": "gulp --gulpfile gulpfile.js test:build"
```

If you pass additional parameters: `npm run prepush -- some_git_parameters`, it'll result in `gulp --gulpfile gulpfile.js test:build some_git_parameters`.

And `gulp` will complain about `some_git_parameters` not being a known tasks.

So, you can't really pass git parameters directly to scripts.

Instead, for `commit-message` and similar hooks, except if I'm missing something, I think it's safe to assume that the commit message can be found `.git/COMMIT_EDITMSG`.

So it should be possible to write:

```js
// precommitmsg.js
var fs = require('fs')
var msg = fs.readFileSync('.git/COMMIT_EDITMSG', 'utf8')
// Modify msg...
fs.writeFileSync('.git/COMMIT_EDITMSG', msg)
```

and in `package.json`:

```json
"precommitmsg": "node precommitmsg.js"
```

deepsweet:
thanks.

jagannath7775:
thanks

PR #34, opened 2015-11-24
Clarify available npm script names

sapegin:
I’ve spent some time to understand that names of Git hooks and npm scripts are not the same.


chicoxyzzy:
makes sense imo

typicode:
Thank you for the table and suggestion :)

PR #35, opened 2015-11-26
Replace `npm rm` with `npm uninstall`

alebelcor:
I find using `npm uninstall` clearer to any of its aliases (`remove`, `rm`, `r`, `un`, `unlink`) mainly because of its counterpart being `npm install`.

Hopefully you'll agree.

Thanks!


iamstarkov:
it is matter of taste, isnt it?

typicode:
I agree. Personally, I'm used to `npm rm` because it's shorter, but for beginners it may be more intuitive to use `npm uninstall`. Also, I verified in npm CLI documentation and they reference the longer version `uninstall` in their docs https://docs.npmjs.com/cli/uninstall.

Thank you for the PR :)

PR #39, opened 2016-02-20
Fix bin scripts to use proper nodejs binary

narqo:
There is no guaranty that `node` binary is present in the user's path during module installation. It's a common practise to use `/usr/bin/env node`, so call `/opt/nodejs/5/bin/node /opt/nodejs/5/bin/npm install` could work.

Unfortunately, I don't have any Windows experience and no possibility to test that. So it'd be awesome if anybody could verify everything still works on Windows.

### DIFF
--- a/bin/install.js
+++ b/bin/install.js
@@ -1,3 +1,5 @@
+#!/usr/bin/env node
+
 // Run when package is installed
 var fs    = require('fs')
 var husky = require('../src/')
--- a/bin/uninstall.js
+++ b/bin/uninstall.js
@@ -1,3 +1,5 @@
+#!/usr/bin/env node
+
 // Run when package is uninstalled
 var husky = require('../src/')
 var hooks = require('../src/hooks.json')
--- a/package.json
+++ b/package.json
@@ -5,8 +5,8 @@
   "main": "index.js",
   "scripts": {
     "test": "node test",
-    "install": "node ./bin/install.js",
-    "uninstall": "node ./bin/uninstall.js"
+    "install": "bin/install.js",
+    "uninstall": "bin/uninstall.js"
   },
   "repository": {
     "type": "git",


typicode:
Hi @narqo,
Thank you for the PR :+1:

typicode:
It's released in `v0.11.2`

narqo:
That's great. Thank you!

narqo:
@deepsweet 

>> So it'd be awesome if anybody could verify everything still works on Windows.

> it fails on windows:

Crap, this is exactly what I was afraid of. 

But I don't understand how other CLI tools (e.g. gulp/babel-cli/etc) work on Windows? I believe I don't call them as `node gulp`, do you? What if `bin/{install.js,uninstall.js}` files will be marked as executable with the "bin" section of `package.json`?

Another possibility is to check how previous version will work with the [upcoming npm@2.14.21](https://github.com/npm/npm/releases/tag/v2.14.21).

> [..] using node is a common practice

I'm still not sure about that. One use `node <file>.js` for things that aren't supposed to be executable, but not for things that are in the `bin/` directory.

BTW @typicode @deepsweet looks like https://appveyor.com/ might be useful for the project.

deepsweet:
"real" bins (declared in `bin` field of `package.json`) will be transformed
into tricky cmd/bat files on Windows by NPM because unix shebangs does
nothing there.

these install/uninstall scripts are normal js files used only by Husky
itself (and that's why they are not exposed to `node_modules/.bin`), so it
should be ok to use `node` here.

typicode:
Fixed, sorry for the bad release I've forgotten about this case.

deepsweet:
:+1:

typicode:
I think for the moment it's better to keep it simple.

Also regarding the original issue, if `npm` is running, actually I can't see how `node` wouldn't be in the `PATH`. Except if someone reports an issue let's keep it the way it is.

You're right appveyor is useful but I feel that the project is simple enough (and won't change that much) to add it for the moment. That said, I'm using it for [hotel](https://github.com/typicode/hotel) and I'm pretty happy with it :).

narqo:
> Also regarding the original issue, if npm is running, actually I can't see how node wouldn't be in the PATH. Except if someone reports an issue let's keep it the way it is.

Actually this is exactly the issue that I was tried to solve. Our CI environments don't have node in the PATH, but we have different node versions installed in `/opt/node/<version>`. Current we had put husky to `optionalDependencies`, as it can't be installed during build. Unfortunately, all this CI machinery can't be fixed in the nearest future :(

PR #45, opened 2016-04-27
Export path environment

nwinch:
When using git with a GUI app (like Tower) instead of the CLI, it [doesn't know about the environment variables](https://www.git-tower.com/help/mac/faq-and-tips/faq/hook-scripts) you may have set up and just ignores any git hooks.

Exporting `PATH` inside the hook script itself allows the GUI to run the hook as expected.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -46,6 +46,7 @@ module.exports = {
 
     // Hook script
     arr = arr.concat([
+      'export PATH=' + process.env.PATH + ':$PATH',
       'cd ' + normalizedPath,
 
       // Fix for issue #16 #24


MadRabbit:
good find! although, did you know you can add things dynamically in `~/.bash_profile`? like for example i have a bin stub like that in mine:

```
export PATH=./bin:./node_modules/.bin:/usr/local/sbin:/usr/local/bin:/usr/local/share/npm/bin:$PATH
```

this thing makes everything visible in a local `bin` and `node_modules/.bin` directories wherever I go. Maybe you could setup a similar thing to make your `pwd` to be automatically visible in your `$PATH`?

I reckon that would solve the problem for you in all the projects at once

nwinch:
Yep, my `PATH` looks similar.  Though as far as I can tell, Tower won't read that?

A snippet from the link I provided in the description:
> In that case, if you want your hook scripts to run in different environments (or just GUI apps), the only reliable way to achieve this is to set up your required environment in the hook script itself.

So in order for Tower to be aware of your existing `PATH` variable, I've loaded it into the hook file.

typicode:
Hi @nwinch,

Thank you for the PR. That's a good idea, husky was even doing this at some point ;)
But some concerns were raised about this approach with `nvm`.

For example, if the user installs husky using Node version X and there's a security fix that is released. The user will upgrade to version Y using `nvm` but the hook script will still use version X.

Or if the user simply uninstalls the version used to install husky, the hook script will fail.

Instead, the current approach is to look for `npm` CLI in paths used by `nvm`, `homebrew` and Node installer. So husky should work if Node was installed using one of these tools.

How did you install Node? Could you run `which npm`?

nwinch:
Ahh I see, that makes sense.

I'm using [n](https://github.com/tj/n) for node version management.  TBH I'm not tied to it, and just now researching the difference between `n` and `nvm`, I'd be happy to switch over anyways :)

Unless you feel it's worth supporting the way `n` manages versions?

nwinch:
Works a treat with `nvm`.  Thanks!

lukepolo:
Is there something we can do for this? Dont mind if its a local change in our repo etc.

Issue #26, opened 2015-08-12
silent npm run


iamstarkov:
why did you choose run npm scripts in silent? usually its hiding some problems

typicode:
Originally, it was to make hooks output less verbose and more readable:

```
$ npm run test
src/actions/daemon.js -> lib/actions/daemon.js
src/actions/servers.js -> lib/actions/servers.js
...
```

vs

```
$ npm run test

> hotel@0.3.7 test /home/typicode/d/hotel
> npm run build && mocha && npm run standard 


> hotel@0.3.7 build /home/typicode/d/hotel
> babel src --out-dir lib --copy-files --loose all

src/actions/daemon.js -> lib/actions/daemon.js
src/actions/servers.js -> lib/actions/servers.js
...
```

Though, I agree that it hides things.

But if you need to debug stuff, you can run `npm run prepush` and you'll get the full output. That's one of the cool thing of having them defined in `package.scripts`.

iamstarkov:
I thought about it sometime, and ended up with thought that its nice to hide if everything is okay, but if it failed, then it will be better to show the error message. what do you think?

scotthovestadt:
> I thought about it sometime, and ended up with thought that its nice to hide if everything is okay, but if it failed, then it will be better to show the error message. what do you think?

+1

iamstarkov:
@typicode any thoughts?

typicode:
Sorry forgot to reply. I'm on vacation with just my phone, will make a better/more detailed answer when I get back :)

andrew-makarenko:
I found this behaviour confusing, so it would be great to add some note about this in the readme.
Smth like 'Hook scripts will be runned with --silent flag, so you wouldn't see any information passed to STDOUT by your script. If you need to debug your script, you can run it manually `npm run prepush`'

pgom:
@typicode actually it would be great if the silent (or other) flag could be used as an option. What do you think?

typicode:
I prefer to avoid adding options to keep it simple and I'm kind of satisfied with husky not being `silent`. Also I'm not sure what other flag could be added?

lili21:
I think optional silent would be helpful. When I am using git hooks for linting source code, those lint errors would be enough.

PR #47, opened 2016-05-07
don't output  npm default error message

ccbikai:
So annoying

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -91,7 +91,7 @@ module.exports = {
       'command -v npm >/dev/null 2>&1 || { echo >&2 "' + npmNotFound + '"; exit 0; }',
 
       // Run script
-      'npm run ' + cmd,
+      'npm run -s ' + cmd,
       'if [ $? -ne 0 ]; then',
       '  echo',
       '  echo "husky - ' + name + ' hook failed (add --no-verify to bypass)"',


typicode:
Thank you for the PR and the feedback.
There were some "debate" about it https://github.com/typicode/husky/issues/26

It's kind of hard to decide what should be the best by default. So in the end, husky output is the same as `npm run <hook-script>`.

But you can configure npm to be more silent:

```sh
project $ cat .npmrc
loglevel=silent
```

all3fox:
Hey @typicode, I have a similar problem with npm flags in jzaefferer/commitplease#60

My goal is to pass the `--silent` flag when the `commitmsg` npm script is run. I do not want to affect npm behvaiour in other cases, so setting verbosity in `.npmrc`, like above, is not an option.

Husky uses bash scripts inside `.git/hooks`. In particluar, the `commit-msg` bash script has the line:
```
npm run commitmsg
```
where `commitmsg` comes from package.json. So, I tried these two approaches in package.json:
 1. `"commitmsg": "--silent node node_modules/commitplease"`
 2. `"commitmsg": "-s node node_modules/commitplease"`

In my opinion, inside the bash script they are supposed to become:
 1. `npm run --silent node node_modules/commitplease`
 2. `npm run -s node node_modules/commitplease`

Instead the hook fails with the messages (the number of dashes differs):
 1. `sh: --: invalid option`
 2. `sh: - : invalid option`

It works silently if I put `"commitmsg": "node node_modules/commitplease"` into package.json and run manually from console (zsh version 5.2):
```
npm run --silent commitmsg
```

It also works if I run the following in console:
```
sh
sh-4.3$ npm run --silent commitmsg
```

Any hints on how I could debug that?

iamstarkov:
@all3fox i dont get your intention, but maybe you can [use `--` to pass arguments to `npm scripts`](https://github.com/npm/npm/pull/5518)

all3fox:
@iamstarkov thank you for the hint, this is what I tried:

 1. `"commitmsg": "-- --silent node node_modules/commitplease"`
 2. `"commitmsg": "-- -s node node_modules/commitplease"`
 3. `"commitmsg": "node node_modules/commitplease -- --silent"`
 4. `"commitmsg": "node node_modules/commitplease -- -s"`

None of the above pass the `--silent` flag to npm, which is what I want. Number 3 and 4 are actually passing the silent flag to commitplease instead of npm, which is not what I want. But I tried that anyway. Numbers 1 and 2 are failing with the same error that in my previous message:

 1. `sh: --: invalid option`
 2. `sh: - : invalid option`

So, I am still looking for a way to pass the `--silent` flag to npm when executing a particular `npm run` script.

iamstarkov:
you can try to pipe all the input to /dev/null, `"commitmsg": "commitplease > /dev/null"`

iamstarkov:
also you dont need to point to node_modules explicitly in you npm scripts, npm is dynamically updating `$path` to node_modules binaries on the fly

all3fox:
> you can try to pipe all the input to /dev/null, "commitmsg": "commitplease > /dev/null"

Piping all the output to `2> /dev/null` pipes the output of commitplease instead of npm:

 1. Here is the output for `"commitmsg": "commitplease"`
screenshot from 2016-06-04 12-14-22

 2. Here is the output with `"commitmsg": "commitplease  2>/dev/null"`
screenshot from 2016-06-04 12-16-17

So, it does the opposite of what I would like to achieve: instead of hiding the output of `npm`, it hides the output of `commitplease`. I have also tried the following with no luck:
  
 1. `"commitmsg" : "2>/dev/null commitplease"`
 2. `"commitmsg" : "2>/dev/null -- commitplease"`
 3. `"commitmsg" : "commitplease -- 2>/dev/null"`

> also you dont need to point to node_modules explicitly in you npm scripts, npm is dynamically updating $path to node_modules binaries on the fly

thank you for that, I read [this](https://docs.npmjs.com/cli/run-script) and [this](https://docs.npmjs.com/files/package.json#bin) sections of the manual to make npm put a commitplease script under `node_modules/.bin`. It now runs as `"commitmsg": "commitplease"`

However, I still did not figure out a way to pass the `--silent` flag to `npm`

PR #51, opened 2016-06-09
Pass sh parameters from hooks into npm scripts

### LINKED ISSUE #50: Pass hook parameters through to scripts
By default, git passes a variety of parameters through to its hooks.  This lets you do things like write a post-checkout hook that knows which branch it switched from and can compare what changed.

Currently, husky strips those parameters out, but it would be handy (and helpful) if the parameters were passed through using syntax for optional parameters to [`npm run`](https://docs.npmjs.com/cli/run-script).

boxfoot:
Fixes #50

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ npm install husky --save-dev
 git commit -m "Keep calm and commit"
 ```
 
-Existing hooks aren't replaced and adding `--no-verify` to your git commands lets you bypass hooks. You can also use [any Git hook](HOOKS.md).
+Existing hooks aren't replaced and adding `--no-verify` to your git commands lets you bypass hooks. You can also use [any Git hook](HOOKS.md).  Optionally include the environment variable `$GIT_PARAMS` in your scripts to access any command-line parameters provided by git.
 
 ## Tips
 
--- a/src/index.js
+++ b/src/index.js
@@ -91,7 +91,7 @@ module.exports = {
       'command -v npm >/dev/null 2>&1 || { echo >&2 "' + npmNotFound + '"; exit 0; }',
 
       // Run script
-      'npm run ' + cmd,
+      'npm run ' + cmd + ' -- $*',
       'if [ $? -ne 0 ]; then',
       '  echo',
       '  echo "husky - ' + name + ' hook failed (add --no-verify to bypass)"',
--- a/src/index.js
+++ b/src/index.js
@@ -91,7 +91,8 @@ module.exports = {
       'command -v npm >/dev/null 2>&1 || { echo >&2 "' + npmNotFound + '"; exit 0; }',
 
       // Run script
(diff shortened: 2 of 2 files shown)


jessepinho:
Thanks for this, @boxfoot! @typicode, can we get this merged in?

typicode:
Sorry for the delay and thanks for the PR.
Actually, there's an edge case.

For example, `pre-push` takes 2 parameters, but if you have something like:

```json
{
  "prepush": "some-test-command"
}
```

And you pass Git params, `some-test-command` will receive 2 params and may fail because `some-test-command` may not accept 2 params.

Another solution that could be explored would be to have an environment variable like `$GIT_PARAMS` and do something like `"prepush": "some-script $GIT_PARAMS"` (not sure that if there's 2 params it will work though).

jessepinho:
@typicode Bash scripts will generally just ignore the extra params, though, without failing. For example, in my case, I'm running a JS executable for the `commitmsg` hook, and I want to grab the commit message temp file out of `process.argv[2]`. If I didn't actually use the params, though, nothing would fail—the params would just be ignored. So I can't imagine this would cause any problems.

typicode:
You can try with `mocha`

```sh
$ mocha .git/COMMIT_EDITMSG 
/Users/typicode/project/.git/COMMIT_EDITMSG:1
```

jessepinho:
Ah, makes sense. I just did a little experiment:

```
// foo.js
console.log(process.argv);
```

```
$ export FOO="foo bar"
$ node foo.js $FOO
[ '.../node',
  '.../foo.js',
  'foo',
  'bar' ]
```

So it looks like your `$GIT_PARAMS` idea would work!

ValeriiVasin:
@typicode @boxfoot  solution with $GIT_PARAMS is more elegant and will not brake things. We applied change from the pull request and have some problems with pre-push already, because we are using [concurrently](https://www.npmjs.com/package/concurrently) package to speedup tests execution.

it basically tries to do `concurrently "npm run test:frontend" "npm run test:backend" "origin" "<my-git-repo>"` and fails.

boxfoot:
OK!  Here's my attempt to use the GIT_PARAMS proposal (which I agree is very elegant!)

jessepinho:
That looks good! 🐑 🇮🇹

typicode:
Thank you 👍

boxfoot:
You're welcome!

ReDrUm:
Can someone please explain how to use/access the `$GIT_PARAMS`?

Within `package.json` I have:
```
{
    "scripts": {
        "postcheckout": "node ./hooks/post-checkout.js $GIT_PARAMS"
    }
}
```
Then within my `post-checkout.js` logging `process.argv` returns an array containing my path to the node executable, and the path to the current script. I was expecting it to also contain the git params for post-checkout which would be two branch refs and the type flag.

PR #54, opened 2016-07-13
Run NPM command silently

jessepinho:
This will suppress the `NPM ERR!` lines if the script exits with a non-zero exit code. (Those lines may be useful in an NPM context, but this is in a Git context, where the `NPM ERR!` lines are just confusing and noisy.)

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -109,7 +109,7 @@ module.exports = {
 
       // Run script
       'export GIT_PARAMS="$*"',
-      'npm run ' + cmd,
+      'npm run ' + cmd + ' --silent',
       'if [ $? -ne 0 ]; then',
       '  echo',
       '  echo "husky - ' + name + ' hook failed (add --no-verify to bypass)"',


iamstarkov:
We discussed it before and decided to prefer explicit output over implicit one

jessepinho:
@iamstarkov Not sure I understand. What do you mean by "implicit" output?

Also, if this isn't a good solution, what do you think of checking for an `NPM_SILENT` env variable when running the command, and adding `--silent` if it's present? i.e., in my package.json, I'd have:

`"commitmsg": "NPM_SILENT=1 ./bin/foo"`

3y3:
Maybe more reasonable:
```sh
. ./.huskyrc

if [ $NPM_SILENT ]; then
  npm run -s command
else
  npm run command
fi
```

jessepinho:
@3y3 sorry for the delay... I don't think I understand your suggestion. `.huskyrc` isn't a config file used by Husky, as far as I'm aware. (I don't see any references to it in the repo.) Where are you suggesting that code would go?

simplesmiler:
Really want it to be either the default or configurable.

jessepinho:
@simplesmiler Looks like this will be fixed by #73, if that gets merged!

typicode:
I've updated husky and the output in `v0.13.0` should be less verbose by default (see https://github.com/typicode/husky/pull/73#issuecomment-274472394)
It's not fully silent either as it lets the user know that a git hook is running and how to debug it if it fails.

You can keep the same log level as before by adding `-q` to your script in `package.json`.

```json
"precommit": "npm run some-script -q"
```

Thanks for the PR and feedback :+1:

PR #56, opened 2016-07-15
Fixing sh syntax

guilouro:
The old version broke on ubuntu 16 to use syntax 'bash' in 'sh'


typicode:
Hi,
Thank you for the PR :)
I'm on vacation right now, so I won't be able to publish a new version before Monday. Please use the previous version until then.

pedrogustavo:
:+1:

typicode:
Closing it, I've merged #57 which does the same thing minus `exit 0` (actually, I didn't see the value to add it explicitly).

Thanks again for the help :)

Issue #49, opened 2016-05-25
I understand this works with GUI's but what about Github's Desktop Client?


lasergoat:
I have installed node with the standard installer.

I have husky working in the CLI but Github Desktop Client acts like there are no scripts to run.

iamstarkov:
maybe it is the problem with github desktop?

typicode:
Hi,

What's the output of `which node`? You're on OS X?

lasergoat:
`/usr/local/bin/node`

typicode:
> I believe that something within the husky generated script skips the entire hook if node isn't found.

It checks if `npm` is available in `$PATH` and fails otherwise. I suspect that `$PATH` may not be the same when the git hook script is run from GitHub Desktop.

If you want verify it, you can run:

```sh
my-project$ echo $PATH

# echo $PATH in some file when pre-commit is called
my-project$ mv .git/hooks/pre-commit .git/hooks/pre-commit.bak
my-project$ echo 'echo $PATH > ../../precommit-path' > .git/hooks/pre-commit

# try to make a commit from GitHub Desktop and check the output
my-project$ cat precommit-path
```

Sorry I'm on Linux right now, so I can't test it.
There may be a setting in GitHub Desktop to edit `PATH`.

typicode:
Quick update, works with `nvm` + `GitHub Desktop`.
Need to test with Node installed differently.

skhilko:
I confirm, husky hooks don't work for GUI clients with Node installed without use of `nvm`. When I use SourceTree to commit something, I see the following error:
`husky - can't find npm in PATH. Skipping precommit script in package.json`
I'm pretty sure it fails the same in GitHub Desktop.

@typicode, you're exactly right:
> I suspect that $PATH may not be the same when the git hook script is run from GitHub Desktop.

After some research I found that GUI apps on mac didn't respect `$PATH` value set for terminal environment.

I've installed node via homebrew. Both `node` and `npm` are installed under `/usr/local/bin`. (as you'd installed them with a standard installer)

`/usr/bin:/bin:/usr/sbin:/sbin` is the `$PATH` value when committing from SourceTree.

I know that this is not really a husky issue, that is simply how mac GUI apps work. However, if we could include `/usr/local/bin` when husky generates the hooks, that would've resolved the issue for many people without the need of finding [the unreliable solution and tinkering with the system](http://apple.stackexchange.com/questions/51677/how-to-set-path-for-finder-launched-applications).

[...]

PR #63, opened 2016-09-03
Fix npm dir not included in system PATH in mac GUI apps

skhilko:
Closes typicode/husky#49

Located `npm` and included its dir on the system path if it's not there.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
     "rimraf": "^2.2.8"
   },
   "dependencies": {
-    "normalize-path": "^1.0.0"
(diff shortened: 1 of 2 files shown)


typicode:
Thank you for the PR 👍 
Actually, I was more thinking about adding the missing paths to PATH environment variable (https://github.com/typicode/husky/commit/931070e541d2852adeb9d325627dfe19ba38f55e).

I've tested it with GitHub client on OS X and it seems to work. Let me know if you still have the issue.

skhilko:
@typicode, I see... Thanks for fixing that! :wink:

PR #70, opened 2016-10-09
Support cygpath for cygwin-based git executable

kwonoj:
This PR updates behavior of getting git path for edge-case windows users, like using git executable from `msys2` or similar (i.e, `cygwin`).

Since those git executable returns path as *nix style in windows system, it'll end up husky hook installation fails by

> husky
> setting up hooks in .git/hooks
> fs.js:922
>   return binding.mkdir(pathModule._makeLong(path),
>                  ^
> 
> Error: ENOENT: no such file or directory, mkdir 'Z:\z\github\g.git\hooks'
>     at Error (native)
>     at Object.fs.mkdirSync (fs.js:922:18)
>     at Object.module.exports.create (Z:\github\g\node_modules\husky\src\index.js:123:33)
>     at Z:\github\g\node_modules\husky\bin\install.js:22:13
>     at Array.forEach (native)
>     at Z:\github\g\node_modules\husky\bin\install.js:20:11
>     at Z:\github\g\node_modules\husky\src\index.js:17:9
>     at ChildProcess.exithandler (child_process.js:193:7)
>     at emitTwo (events.js:106:13)
>     at ChildProcess.emit (events.js:191:7)

try to create hook under `Z:\z\github\g\.git\hooks`, instead of correct windows path `Z:\github\g\.git\hooks`

In changes, it borrows `cygpath` included in `msys2` and `cygwin` and let it correct path to windows style paths.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
   },
   "dependencies": {
     "is-ci": "^1.0.9",
-    "normalize-path": "^1.0.0"
+    "normalize-path": "^1.0.0",
+    "which": "^1.2.11"
   }
 }
--- a/src/index.js
+++ b/src/index.js
@@ -2,6 +2,7 @@ var fs = require('fs')
 var path = require('path')
 var exec = require('child_process').exec
 var normalize = require('normalize-path')
+var which = require('which')
 
 module.exports = {
   isHusky: function (filename) {
@@ -14,7 +15,18 @@ module.exports = {
       if (error) {
         callback(stderr, null)
       } else {
-        callback(null, stdout.trim() + '/hooks')
+        var hooksPath = stdout.trim() + '/hooks'
+        if (process.platform === 'win32' && which.sync('cygpath')) {
+          exec('cygpath -w ' + hooksPath, function (cygError, cygStdout, cygStdErr) {
+            if (cygError) {
+              callback(cygStdEr, null)
+            } else {
+              callback(null, cygStdout.trim())
+            }
+          })
+        } else {
+          callback(null, hooksPath)
+        }
       }
     })
   },
--- a/test/index.js
+++ b/test/index.js
@@ -3,13 +3,14 @@ var fs = require('fs')
 var path = require('path')
 var rmrf = require('rimraf')
 var husky = require('../src/')
+var normalize = require('normalize-path')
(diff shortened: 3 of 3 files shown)


the0neWhoKnocks:
I was literally just working on a fix for this. My fix was
```
// src/index.js:17

if( stdout.indexOf('cygdrive') > -1 ){
  stdout = stdout.replace(/\/cygdrive\/([a-z])\//, '$1:/')
}
```

Yours seems more comprehensive. Any idea when this'll make it in? Would be nice to be able to use this with Cygwin.

kwonoj:
@the0neWhoKnocks suggested regex will work on `cygwin`, but won't on `msys2`. I'd thought of direct string match-replace first, but choose to rely on cygpath instead.

the0neWhoKnocks:
@kwonoj - correct hence my "more comprehensive" statement ;-). 

@typicode - There an idea of when this'll get merged in?

typicode:
Since cygwin doesn't seem to have a great support in Node, I have mixed feelings about adding cygwin specific commands/code in husky.
https://github.com/nodejs/node-v0.x-archive/issues/6459
https://github.com/npm/npm#installing-on-cygwin

That said, I'm considering https://www.npmjs.com/package/find-parent-dir as an alternative way to find `.git` directory than `git rev-parse`.

I've not yet tested it on cygwin but maybe it wouldn't have the same issue with `Z:\z\github\.git\hooks`?

kwonoj:
I've been previously suggested `app-root-path` module in other packages having similar purpose (https://github.com/okonet/lint-staged/pull/19) which works gracefully on most windows configuration. This PR intended to amend existing behavior only, if you'd update behavior to not to use git it's fine to close PR without check in.

typicode:
I've just published a beta version that doesn't rely on git command and should work with cygwin. If you want to give it a try:

```
npm install husky@beta 
```

Let me know if you still have issues with it.

the0neWhoKnocks:
@typicode - Gave the beta a whirl and things are working as they used to again. :+1:

typicode:
Awesome thanks for the feedback and help.

aminya:
Isn't this related to #673 ?

Issue #64, opened 2016-09-11
Seems not respecting `--silent`


tunnckoCore:
It seems it not respects the `alias npm="npm --silent"`.

iamstarkov:
what do you mean by that?

tunnckoCore:
i have this alias, and when fails for example the commit-msg hook it outputs npm's 20 lines error, which i dont like and that's silent flag for

tunnckoCore:
i believe, it's because you use bash's `command npm ...`.

i have an idea, why not just call `node path/to/jsfile.js` instead of creating and workin on bash file. i'll show what i mean, im on phone now.

iamstarkov:
husky is separated from npm, so how npm silent flag will help you with husky?

tunnckoCore:
What about if we just create simple Bash file for in the hook dir, then just call node from it.

**example .git/hooks/commit-msg**

```bash
#!/usr/bin/env bash
node path/to/file.js
```

And actually the hook will be plain javascript file which will allow cool things. Idea is more advanced and I'm going to implement it in https://github.com/tunnckoCore/npm-hooks

iamstarkov:
npm hooks' environment is different from plain node environment

mrmckeb:
It would be great to resolve this, having large NPM errors logged out makes the actual error harder to see.

marvinhagemeister:
PR #73 is a different config based approach which would satisfy both positions. At work there are some of us who prefer to suppress npm's verbose error output and some who prefer the verbose version.

marvinhagemeister:
We'll my coworkers were bugging me a lot with this the past week, so I finally gave in and created a minimal fork with my PR applied. It is published on npm under the package name: [husky-dontbark](https://github.com/marvinhagemeister/husky/tree/husky-dontbark).

Obviously a fork is not an optimal solution and I'd love to bring some sort of silence mode into core (either via my PR or something else). We basically have two different opinions in this thread: The first group who expects errors to be explicit and the second one who trusts the actual scripts that get called to output anything useful.

In either case husky itself prints a sweet error message upon a failing hook:

```bash
husky - pre-commit hook failed (add --no-verify to bypass)
```

davidhund:
@typicode: any plans on merging this PR #73?

tunnckoCore:
:+1:

nerdgore:
Has anyone tried an npmrc file? It works wonderfully for me. IMO there is no need to add a config option to the package.json.
[...]

PR #73, opened 2016-10-15
Allow custom husky config

marvinhagemeister:
This PR allows the user to pass custom configuration options to husky via the (little known) [config-property](https://docs.npmjs.com/files/package.json#config) inside `package.json`.

In a nutshell this is a different approach to https://github.com/typicode/husky/pull/54, with the addition that the current defaults remain the same and we can satisfy both worlds.

It works by getting traversing the tree upwards until it reaches the topmost `package.json` and using the `config.husky` property if any is present. Otherwise it simply falls back to an empty config.

``` json
{
  "name": "mypackage",
  "scripts": {
    "precommit": "whatever..."
  },
  "config": {
    "husky": {
      "silent": "true"
    }
  }
}
```

Fixes https://github.com/typicode/husky/issues/64.

### DIFF
--- a/bin/install.js
+++ b/bin/install.js
@@ -1,8 +1,11 @@
 // Run when package is installed
-var fs    = require('fs')
-var isCI  = require('is-ci')
-var husky = require('../src/')
-var hooks = require('../src/hooks.json')
+var fs       = require('fs')
+var isCI     = require('is-ci')
+var findRoot = require('find-root')
+var husky    = require('../src/')
+var hooks    = require('../src/hooks.json')
+
+var package  = findRoot('package.json')
 
 if (isCI) {
   console.log('\033[4;36m%s\033[0m', 'husky')
@@ -17,9 +20,13 @@ husky.hooksDir(function(err, dir) {
   if (err) {
     console.error('  ' + err)
   } else {
+    var config = package && package.config && package.config.husky
+      ? package.config.husky
+      : {};
+
     hooks.forEach(function (hook) {
       script = hook.replace(/-/g, '')
-      husky.create(dir, hook, script)
+      husky.create(dir, hook, script, config)
     })
     console.log('done\n')
   }
--- a/bin/install.js
+++ b/bin/install.js
@@ -1,11 +1,9 @@
 // Run when package is installed
 var fs       = require('fs')
 var isCI     = require('is-ci')
-var findRoot = require('find-root')
+var path     = require('path')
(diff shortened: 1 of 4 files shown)


marvinhagemeister:
Oh well forgot to check for root `package.json`. Fix is coming

noamokman:
check out [cosmiconfig](https://github.com/davidtheclark/cosmiconfig) 😄

marvinhagemeister:
@noamokman Well, that seems to be totally overkill for what we're doing here.

noamokman:
@marvinhagemeister it just saves you the entire `getPackageJson` function 😅 Good luck!

marvinhagemeister:
@noamokman I'd much rather use 20 LOC than introduce a new dependency that is more like 500LOC. The awesome thing about husky is that it is slim without any additional baggage like most other tools in the npm world.

mrmckeb:
Is this PR still in review? I'd love for this to be merged in.

jessepinho:
*bump*

Could we get an update on this from the maintainers? There are now conflicts due to how long this has sat here. I'd be happy to resolve them in a branch, but would like to first make sure that it will actually be reviewed by the maintainers and merged...?

typicode:
Hi all,

Sorry for the delay and thanks for the PR.

If possible, I prefer to keep husky "zero-config".

That said, since there was a lot of feedback/requests about it, in `v0.13.0`, scripts will be run with `-s` but some additional messages will be output to let the user know that a Git hook is running.

Overall `v0.13.0` should be less verbose, especially when a script fails :)

marvinhagemeister:
@typicode Awesome! That is the best of both worlds!

xRoulanDx:
[WOW WOW WOW](https://github.com/typicode/husky/issues/26#issuecomment-157158853)

Looks like we have two sides of users of this package. I don't want to run my precommit script for detailed information about why my pre-commit hook is failed. I want to look at the output git log and get an info about what is wrong with my script. We need an ability to configure the silent attribute. Like plz :)

Ghostbird:
I think there's a clear argument for a way to pass `--silent` to `npm` here. Some users have a legitimate need to see all errors, but I'm gettin 19 lines of errors. Only line 9 and 10 are the actual details of where and why my linter failed. All the rest is boilerplate that will be early the same and almost always unimportant.

So I have to fish my actual error message out of a sea of generic noise.
`NPM ERR! code 1
...`

Issue #20, opened 2015-06-04
0.8.1 breaks on a git submodule


avaly:
```
$ git commit -m "Test"
/src/web-dev/.git/modules/web-tools/hooks/pre-commit: 5: cd: can't cd to ../../web-tools
```

Where the project structure is:
- `/src/web-dev` - git repo with multiple git submodules
- `/src/web-dev/web-tools` - one of such git submodules which contains a `package.json` and where husky is installed in `node_modules`

Shutnik:
Same problem, but works well without [this](https://github.com/typicode/husky/commit/62ca38fb0f51b2106ad2ba0bafe3ac4366ded0cc) changes.

IcanDivideBy0:
any updates on this ? does someons found a fix / workaround ?

PR #83, opened 2016-12-03
Fixes issue with submodules with a package.json not in .git's dir

jrobichaud:
This is a fix for issue #20 

In a submodule (with git version 1.7.8+) "dir" is located inside parent's .git directory.

`var projectDir = path.join(dir, '..', '..')` is not the appropriate folder in this case.

How I implemented it is that I calculate the path from the hook directory to the `package.json`'s directory which is accurate all the time.

The hook script moves from its folder to the relative path provided.

Another approach would be to run both `git rev-parse --show-toplevel` and `git rev-parse --git-dir` then calculate a path satisfying all cases.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -34,19 +34,16 @@ module.exports = {
     // Assuming that this file is in node_modules/husky/src
     var packageDir = path.join(__dirname, '..', '..', '..')
 
-    // dir being .git/hooks
-    var projectDir = path.join(dir, '..', '..')
-
     // In order to support projects with package.json in a different directory
-    // than .git, find relative path from project directory to package.json
-    var relativePath = path.join('.', path.relative(projectDir, packageDir))
+    // than .git, find relative path from hook directory to package.json
+    var relativePath = path.join('.', path.relative(dir, packageDir))
 
     // On Windows normalize path (i.e. convert \ to /)
-    var normalizedPath = normalize(relativePath)
+    var normalizedPath = normalize('/'+relativePath)
 
     // Hook script
     arr = arr.concat([
-      'cd ' + normalizedPath,
+      'cd $(dirname "$0")' + normalizedPath,
 
       // Fix for issue #16 #24
       // Test if script is defined in package.json


jrobichaud:
It appears it does not work on windows, the second approach has to be used.

typicode:
Hi @jrobichaud,

Thanks for the PR. I'm working on a different approach currently.

The idea is to use Node only to figure out the path:
https://github.com/typicode/husky/blob/next/src/index.js#L24

I hope this way it should more reliable across platforms.

I need to clean code a bit and test it on the different platforms.
For submodules, it works by parsing the `.git` file that you can find in a submodule directory.

typicode:
Just published a version with submodule support, if you want to give it a try:

```
npm install husky@beta
```

Feedback welcome :)

jrobichaud:
Ok, ill test it right away.

jrobichaud:
version 0.12.0-1 is not working

The scripts have a "cd" from the parent's git to the package.json location.

However If i manually add a "pwd" in a post-checkout script the current directory is the location of the ".git" file of the submodule is.

typicode:
Thanks. Do you have this kind of organization?

```sh
/.git
/some/path  # submodule directory
/some/path/.git
/some/path/package.json
```

For the `cd`, it's something like that `cd some/path` or is it something else?

typicode:
Also, what should `cd` be?

jrobichaud:
close, our package.json is in a subdir.

/.git
/a/b #submodule directory
/a/b/.git
/a/b/c/package.json

cd a/b/c

pwd is /Users/myself/Documents/git/a/b/

it should be "cd c"

jrobichaud:
Feel free to ask me anytime to test, it takes only a few seconds each time.

typicode:
Thank you :) I've updated tests and code to support this kind of layout:
https://github.com/typicode/husky/blob/next/test/index.js#L70

And a new `husky@beta` version (`v0.12.0-2`) has been published.

PR #90, opened 2017-01-04
add npm run cmd -s support

dushaobindoudou:
add npm run cmd -s support


typicode:
@dushaobindoudou Thanks for the PR.

I think just `-s` is a little too silent but at the same time there was a lot of feedback that `npm run` was too verbose.

So in `v0.13.0`, husky is a mix, it uses `-s` but outputs some additional messages too. Overall it should be less verbose :)

PR #91, opened 2017-01-06
Support for SmartGit on OSX

ben-poole:
When running SmartGit v6 on OSX v10.11.2 process.platform is set to 'linux'.


typicode:
Closing as `1.0` may not suffer from this issue.

PR #100, opened 2017-02-21
nvm loading fixed

marudor:
This fixes nvm loading.
it tried to read NVM_DIR/nvm.sh, it correctly searches for the nvm.sh now.
It also tried exists which is not defined. I guess command_exists was intended.


rainboxx:
@marudor This should be fixed in the newest version due to #101.

typicode:
Hi @marudor,

Thanks a lot for the PR.
Sorry, didn't see it before [#101](https://github.com/typicode/husky/pull/101).

PR #101, opened 2017-02-24
Fix loading `nvm`

rainboxx:
`nvm` couldn't be loaded when a specific version of `npm` was in the path even when a `.nvmrc` was added to the project root. This change allows to place a file `.nvmrc` into the project root which will be picked up by husky.

Additionally two apparent bugs are fixed:
* `exists` doesn't seem to be a command neither on macOS nor on Linux, might be a typo
* the test for the file `nvm.sh` is wrong

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -58,8 +58,8 @@ function getHookScript (hookName, relativePath, cmd) {
 
     'load_nvm () {',
     '  export $1=$2',
-    '  [ -s "$1/nvm.sh" ] && . $1/nvm.sh',
-    '  exists nvm && [ -f .nvmrc ] && nvm use',
+    '  [ -s "$2/nvm.sh" ] && . $2/nvm.sh',
+    '  command_exists nvm && [ -f .nvmrc ] && nvm use',
     '}',
     '',
 


rainboxx:
It should be discussed whether this is the intended behavior.

On the one hand I'd say that if a `.nvmrc` is in the project it should also be used before anything else. The current implementation with this PR would load `nvm` even when there's `npm` in the `PATH` and no `.nvmrc` in the project root. Maybe the logic should be 1) check for `.nvmrc`, then try to load `nvm`.

On the other hand it looks like it was me installing node without nvm some time ago and it's not shipped with macOS. So basically it's my fault (except maybe the typos).

I'd still argue that if a project has a `.nvmrc` it should be used for git hooks.

What is your opinion @typicode? I'm happy to change the logic so that it first checks for a `nvmrc` file prior to loading `nvm`.

typicode:
Hi @rainboxx,

Thank you for the PR. Good question, would need to think a little more about this case.
Would you mind changing the PR with only the fix for `exists`? This way it could be merged now.

rainboxx:
@typicode of course. I amended the commit to only fix `exists` and the parameters.

If you'd like to discuss use cases for loading via `nvm` even when `npm` is in the path I'm happy to help.

typicode:
Thank you :)

rainboxx:
Well, thank you for taking this PR :-).

PR #104, opened 2017-03-01
Fix package entry point

ooflorent:
Before this changes, the entry point was erroneous causing errors when requiring this package:

```js
const husky = require("husky")
```
```
Error: Cannot find module 'husky'
    at Function.Module._resolveFilename (module.js:470:15)
    at Function.Module._load (module.js:418:25)
    at Module.require (module.js:498:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (/Users/fc/husky-test/scripts/install.js:3:15)
    at Module._compile (module.js:571:32)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:488:32)
    at tryModuleLoad (module.js:447:12)
    at Function.Module._load (module.js:439:3)
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "husky",
   "version": "0.13.1",
   "description": "Prevents bad commit or push (git hooks, pre-commit/precommit, pre-push/prepush, post-merge/postmerge and all that stuff...)",
-  "main": "index.js",
+  "main": "./src/index.js",
   "scripts": {
     "test": "mocha && standard",
     "precommit": "npm test",


typicode:
Thank you, I didn't think of this usage actually.
Are you using the project as a module?

ooflorent:
> Are you using the project as a module?

Yes. We have an `install.js` script to setup our dev environment. I want it to force the installation of the git hooks using the programmatic API of `husky`.

While currently it's not ideal, it's better than nothing!

```js
const husky = require("husky")

function installGitHooks() {
  husky.installFrom(
    path.dirname(require.resolve("husky/package.json"))
  )
}
```

Here are a few things I would like to change:

- output should be delegated to the CLI tool
- public API should not require husky's path (auto-discover)
- git hooks generation could use options / flags to be configurable (silent mode)

Are you ok if I submit separate PRs to address these changes?

typicode:
> We have an install.js script to setup our dev environment

But isn't it redundant with the fact that `husky` installs itself?

> git hooks generation could use options / flags to be configurable (silent mode)

I prefer to keep it zero-config but with sane defaults (that's why feedbacks are very important).
Regarding silent mode, have you updated to `v0.13`? It should be a lot more silent.

typicode:
Currently, I'm thinking about making a CLI to be added to `scripts > install` in `package.json` instead.
Regarding the refactor and supporting officially a public API, I would need to think about it.

Thanks again :)

Fun fact, just released the previous week a package to prevent this kind of issues [pkg-ok](https://github.com/typicode/pkg-ok) :)

ooflorent:
Thanks!

PR #107, opened 2017-03-07
update failure message

Judahmeek:
The current failure message kind of threw me off when I first saw it.


typicode:
Thank you, missed it 👍

PR #114, opened 2017-03-28
Correct work with worktrees

alexeyten:
Worktree's `.git` file contains absolute path to repository.
`path.join` incorrectly concatenates `/path/to/worktree`
with `/path/to/real/.git/worktrees/worktree` to
`/path/to/worktree/path/to/real/.git/worktrees/worktree`
instead of simple `/path/to/real/...`

Using `path.resolve` instead of `path.join` resolves this problem
while relative paths of submodules still works correctly.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -35,7 +35,7 @@ function findHooksDir (dirname) {
         .split(':')[1]
         .trim()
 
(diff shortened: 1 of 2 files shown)


typicode:
Thank you!

PR #126, opened 2017-05-08
Add yarn support (closes #102)

### LINKED ISSUE #102: PR suggestion: add yarn-or-npm as dependency
First of all, thanks to @typicode and other contributors for husky! :bowing_man: 
I discovered [yarn-or-npm](https://github.com/camacho/yarn-or-npm) and I think it suits perfectly for husky. Should I add a PR?

onbjerg:
Husky will now check if yarn is installed, and if it is, then it will run the scripts through yarn instead of npm.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -2,6 +2,7 @@ var fs = require('fs')
 var path = require('path')
 var normalize = require('normalize-path')
 var findParentDir = require('find-parent-dir')
+var yarnOrNpm = require('yarn-or-npm')
 var hooks = require('./hooks.json')
 var pkg = require('../package.json')
 
@@ -50,6 +51,9 @@ function findHooksDir (dirname) {
 function getHookScript (hookName, relativePath, cmd) {
   // On Windows normalize path (i.e. convert \ to /)
   var normalizedPath = normalize(relativePath)
+
+  // Check if we're using yarn or npm
+  var packageManager = yarnOrNpm()
 
   // Hook script
   var arr = [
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -123,30 +127,30 @@ function getHookScript (hookName, relativePath, cmd) {
     ])
   }
 
-  // Can't find npm message
-  var npmNotFound = '> husky - Can\'t find npm in PATH. Skipping ' + cmd + ' script in package.json'
+  // Can't find package manager message
+  var pmNotFound = '> husky - Can\'t find ' + packageManager + ' in PATH. Skipping ' + cmd + ' script in package.json'
```
faddee: Maybe it should say "Can't find yarn or npm in PATH"?

onbjerg: Probably. Right now it will almost always say "Can't find npm in PATH" since it only checks if yarn exists if it's already in the PATH (because we only use yarn if we detected it in PATH already).

I'll fix.

victornoel:
will this be merged? :) thanks

GAumala:
There's already a bash function to check if a command exists. I think it would be better to use that rather than adding an extra node module that spawns a child process just to perform this check.

I have implemented a bash only solution in my fork. If you guys are interested you can find it here: https://github.com/GAumala/husky/commit/1d9b8d876282ed03a8b862672fea947ac3989424

fernandopasik:
This would be a great addition, nowadays you can have node without npm and only have yarn installed.

fernandopasik:
In addition to my previous comment when you install node with brew it doesn't come with npm.

@typicode is this improvement not ok for husky?

onbjerg:
>In addition to my previous comment when you install node with brew it doesn't come with npm.

Interesting if true, because npm is still installed with node for me using homebrew.

fernandopasik:
@onbjerg the current node brew formula uses the flag --without-npm by default

fernandopasik:
@typicode would it be possible to reconsider having support for yarn?

typicode:
Hi @fernandopasik,
Yes, working on a new version and it's being reconsidered :)

fernandopasik:
@typicode thanks!!

arsduo:
Hi! We just started using Husky in a project that uses yarn. The beta documentation doesn't mention yarn -- is that still under consideration?

typicode:
@arsduo yes, even though it's not documented (just to keep doc "simple"), yarn should work out of the box. Actually, it should work better than before as husky v0.15 supports also environment where only yarn is installed and not npm.

PR #147, opened 2017-06-28
Fix. Escape spaces in window file path

yurasid:
Spaces in file path for windows are not escaped.

So in the hook file we have something like this:
`cd Solution/05. Web/Service.Web
`

Should be: 
`cd Solution/05.\ Web/Service.Web
`

### DIFF
--- a/__tests__/index.js
+++ b/__tests__/index.js
@@ -45,7 +45,7 @@ describe('husky', () => {
     const hook = readFile(dir, '.git/hooks/pre-commit')
 
     expect(hook).toMatch('#husky')
-    expect(hook).toMatch('cd .')
+    expect(hook).toMatch('cd "."')
     expect(hook).toMatch('npm run -s precommit')
     expect(hook).toMatch('--no-verify')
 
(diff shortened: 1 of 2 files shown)


typicode:
@yurasid Thank you for the PR

Could you try replacing `cd ${normalizedPath}` with `cd "${normalizedPath}"` instead?

I think it should cover more cases and the RegExp could be avoided
https://unix.stackexchange.com/questions/108635/why-i-cant-escape-spaces-on-a-bash-script

Let me know if it doesn't work :)

yurasid:
Yes, this was my first suggestion. It works ok when I "quoted" all strings in my hook files. But it seems that CI not so happy about that. 
I'll see how can i fix this on weekend. 

To be honest, this is a issue for `normalize-path` package. ))

typicode:
> CI not so happy about that.

I've checked it, seems like tests would need to be updated from `cd some/path` to `cd "some/path"`
https://github.com/typicode/husky/blob/master/__tests__/index.js#L48

> To be honest, this is a issue for normalize-path package. ))

Maybe, I don't know if the author wants to handle this case too.

> I'll see how can i fix this on weekend.

I can spare some time and do it tomorrow, if you want :)

yurasid:
Updated. That was simpler than I thought.

PR #150, opened 2017-07-04
Fixed PATH export

kamil7x:
When using Husky inside WSL, there was a problem with spaces in Windows paths:
```
.git/hooks/pre-commit: 31: export: (x86)/NVIDIA: bad variable name
```

This commit fixes it by adding double quotes for PATH export.


typicode:
Hi @kamil7x,

Thank you, what is WSL?

typicode:
It's published :)

kamil7x:
WSL is Windows Subsystem for Linux (weird order) - it is something like Ubuntu inside Windows so you can use Bash and Linux tools, while still using Windows, really nice for developers, but still has some issues.

Thanks

Issue #117, opened 2017-04-11
Hooks fail if user directory has apostrophe in it


luke-oneill:
My user directory has an apostrophe in it (I know... not a good idea, I've found similar bugs in many different scripts). This means that any hooks fail with the following message:

```
.git/hooks/pre-commit: line 39: syntax error near unexpected token `('
.git/hooks/pre-commit: line 39: `#  echo "> husky - pre-commit hook failed (add --no-verify to bypass)"'
```

The problem line in question is actually line 24 in `pre-commit`:
```
load_nvm NVM_DIR /Users/luke.o'neill/.nvm
```

I can fix manually but changing it to:
```
load_nvm NVM_DIR /Users/luke.o\'neill/.nvm
```

but I need to do this after each time I clear my node_modules and run npm install again.

typicode:
Hi @luke-oneill,

I think husky `v0.14.2` should fix it, thanks to @yurasid's PR.
Let me know if not :)

PR #152, opened 2017-07-14
Support paths which have quotes in them.

luke-oneill:
Fixes the issue described in https://github.com/typicode/husky/issues/117 where paths with a quote in them will cause the githooks to fail when they run.

(Previous solution used `normalize-path` which only normalizes slashes, so added extra function for escaping quotes.)

### DIFF
--- a/__tests__/index.js
+++ b/__tests__/index.js
@@ -157,4 +157,20 @@ describe('husky', () => {
     expect(hook).toMatch('husky')
     expect(hook).not.toMatch('ghooks')
   })
+
+  it('should escape apostrophes in the path', () => {
+    mkdir(dir, '.git/hooks')
+    mkdir(dir, 'node_modules/husky')
+
+    // Override home to be a path with an apostrophe in it
+    process.env.HOME = '/User/apotr\'ophe'
+
+    install(dir, 'node_modules/husky')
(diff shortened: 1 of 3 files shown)


typicode:
There should be no need for that, at least for single quotes.

Paths have now `"` around them:
https://github.com/typicode/husky/blob/master/src/utils/get-hook-script.js#L89

Which makes escaping unnecessary. For example, `cd "/some path/with'single'quotes"` should workd.

luke-oneill:
It is still a problem because the error doesn't occur at the `cd` line (89), it occurs on lines 30 and 49 which do not have quotes around them. That may be a cleaner way of doing it though, without needing another library.

lsuarez-tiempo-dev:
hi!!!! 
I have a space in my username.
Event though I added qoutes by overwritting $home on my .zshrc I still get error messages when I run commit.  Is there any ways to fix this?

typicode:
Hi @lsuarez-tiempo-dev 

I should have closed this PR as it relates to `v0.14`, sorry for the confusion.
Can you try with `v1.0.0-rc.13` (see https://github.com/typicode/husky for install instruction)?

Feel free to create an issue if you still have this problem with `v1.0.0-rc.13`.

PR #163, opened 2017-07-31
add the space between 'skipping' word and hookName

monkindey:
If not, the info will be `skippingpost-commit hook (existing user hook)`,

It will be nice like that `skipping post-commit hook (existing user hook)`


typicode:
Good catch, thank you :)

monkindey:
My pleasure 😀

PR #166, opened 2017-08-04
Allow stdin during hooks

jesstelford:
Enabling interactive prompts in hooks

For more info on interactive hooks, check out [this SO question](https://stackoverflow.com/questions/3417896/how-do-i-prompt-the-user-from-within-a-commit-msg-hook).

In particular, my use-case:

We are currently running tests on `prepush`. We have a quick set of tests, and a longer set of tests. We want to prompt the developer as to which set of tests get run allowing them to move fast on feature branches, but execute the entire test suite at key milestones.

### DIFF
--- a/src/utils/get-hook-script.js
+++ b/src/utils/get-hook-script.js
@@ -109,6 +109,9 @@ module.exports = function getHookScript(hookName, relativePath, npmScriptName) {
       echo "husky > npm run -s ${npmScriptName} (node \`node -v\`)"
       echo
 
+      # Allows us to read user input below, assigns stdin to keyboard
+      exec < /dev/tty
+
       npm run -s ${npmScriptName} || {
         echo
         echo "husky > ${hookName} hook failed ${noVerifyMessage}"


zephraph:
I have this exact issue. For my use case we expect our commit messages to be in a particular format 80% of the time. I want to have a prompt to allow our devs to continue with the commit if the deviation was intentional.

typicode:
Hi @jesstelford 

While it works from the command line, I'm having issues when committing from a GUI (GitHub Desktop):
`.git/hooks/pre-commit: line 33: /dev/tty: No such device or address'`

jesstelford:
Interesting! I honestly have no idea how I'd interface with a GUI... Do you have any tips on where to get started so I can add support?

TxHawks:
This would be a very useful feature for me as well, but I don't think there is any way to make it work for gui clients that don't expose `/dev/tty`

src/utils/get-hook-script.js
```diff
@@ -109,6 +109,9 @@ module.exports = function getHookScript(hookName, relativePath, npmScriptName) {
       echo "husky > npm run -s ${npmScriptName} (node \`node -v\`)"
       echo
 
+      # Allows us to read user input below, assigns stdin to keyboard
+      exec < /dev/tty
```
TxHawks: Maybe an environment variable can be a temporarily workaround for the [gui clients issue](https://github.com/typicode/husky/pull/166#issuecomment-326104859)?

something like 
```sh
if [ "$HUSKY_ENABLE_STDIN" = true ]; then
  exec < /dev/tty
fi
```

This way it's a very clear opt in, and adding something like this to `package.json` should work:
```json
{
  "scripts": {
    "precommit": "cross-env HUSKY_ENABLE_STDIN=true node /some/script"
  }
}
```

I think there isn't really any other way around this, and this is a very valid use-case for teams that exclusively use the command line. 
Further move, `HUSKY_ENABLE_STDIN` should be available under `process.env`, so easy for hook devlopers to provide non-interactive fallback.

willclarktech:
Would love to see this even if it was just for CLI. My use case: asking the developer to specify what sort of commit is being created from a list and prepending a relevant emoji to the commit message.

foisonocean:
Any updates?

SnidelyWhiplash:
Can this get merged if @TxHawks's changes for the GUI are implemented..? I want to use this functionality for prompting the user to bump the version during `pre-push` using https://github.com/BigstickCarpet/version-bump-prompt

typicode:
Will be added in https://github.com/typicode/husky/pull/415

PR #173, opened 2017-08-26
Add sendemail-validate githook

ohbarye:
Hello @typicode , and thank you for the great library.

I found [`sendemail-validate`](https://git-scm.com/docs/githooks#_sendemail_validate) hook is missing, and then added.

The hook seems to be added recently.
ref: https://www.spinics.net/lists/git/msg303675.html

I appreciate you if you would check this PR when you have time.


typicode:
Hi @ohbarye,

Interesting didn't know about this one. Thank you for the PR :)

ohbarye:
Thank you for your review too!

PR #178, opened 2017-09-02
added HUSKY_SKIP_INSTALL env variable to skip hook install

### LINKED ISSUE #172: env variable to prevent install
Hi,
Can we add an env variable to disable the install script?
something like the CI detection.

Let me know if I should open a PR.
Thanks

noamokman:
Hi,

This closes #172.
Let me know if I need to change something.

Thanks!

### DIFF
--- a/bin/install.js
+++ b/bin/install.js
@@ -12,6 +12,11 @@ if (isCI && !process.env.HUSKY_IGNORE_CI) {
   process.exit(0)
 }
 
+if (process.env.HUSKY_SKIP_INSTALL) {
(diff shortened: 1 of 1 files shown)


agconti:
@noamokman thanks for making this PR!

typicode:
Thank you :+1:

satazor:
@typicode Could we have a release on the 0.14.x line with this change? Thank you.

szimek:
I had no idea that such option is present, so I just added `husky` to `optionalDependencies` - this way `yarn/npm install` won't fail if installing `husky` fails. Probably not as good as built-in option, but solves the problem for people still on 0.14.x.

PR #181, opened 2017-09-04
Add scope feature

neotech-starassov:
Hello. I created scope feature for husky.

It's really uber feature for our team. 
We're working on different application layers (different directories) in single repository.
With these feature our team can install husky and limit git hooks triggering scope.

Review is required.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,29 @@ Please use `yarn` `v0.24+`
 
 </details>
 
+## Configuration
+
+### Scope
+
+The scope parameter is optional.
+Scope is a path relative to package.json directory to restrict husky hooks triggering.
+
+### package.json
+
+```javascript
+"husky": {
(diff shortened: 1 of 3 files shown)


typicode:
Hi @neotech-starassov,

Thank you for the PR :)
If I want to test it, what kind of directory layout should I create? Or do you have a scenario for this feature?

neotech-starassov:
I think PR must be closed :) Just realized that `scope feature` is usable only for pre-commit case and can be implemented much more simpler way:
```
    // package.json
    ...
    "precommit": "npm run -s is_changed && lint-staged",
    "is_changed": "[ $((git diff HEAD --name-only; git ls-files -o --exclude-standard --modified) | grep '^src' | wc -l) -ne 0 ]",
    ...

```

Issue #192, opened 2017-10-07
Proposal: Add advice on how to use commit message hook


pepkaro:
In #28, a proposal was made to pass the commit message to the specified script. It was ultimately rolled back for good reasons, but it still meant I didn't have a good idea on how to make my husky commit message hook work.

I ended up doing a bit of searching to figure out how to make my commit message hook meaningful. It might be nice to add an "advice" file or a section to HOOKS.md to help guide users on how to make hooks in corner cases like this.

pepkaro:
If this sounds useful, I added a Pull request (#193) that adds this to HOOKS.md .

PR #193, opened 2017-10-07
#192 - add advice on how to do a commit-msg hook with Husky

pepkaro:
I've added some advice to help others on how to make a commit message hook with Husky. Advice welcome!

### DIFF
--- a/HOOKS.md
+++ b/HOOKS.md
@@ -30,7 +30,7 @@ Advice on how to use specific hooks.
 
 ### commit-msg
 
-Scripts that want to hoook into `commit-msg` must read the commit message themselves; it is not passed in `process.argv`. Instead, the hook script must load the commit message itself. For example, a script could check the message by reading `.git/COMMIT_EDITMSG`. For example, with node,
+Unlike usual with the `commit-msg` git hook, Husky does not pass the commit message in to the `commitmsg` script. Scripts that want to examine the commit message must read the commit message itself. This can be done by reading `.git/COMMIT_EDITMSG`. For example, with node,
 
     var vs = require('fs');
 
--- a/HOOKS.md
+++ b/HOOKS.md
@@ -23,3 +23,16 @@ Husky supports all git hooks (https://git-scm.com/docs/githooks). Simply add the
 | push-to-checkout | pushtocheckout |
 | update | update |
 | sendemail-validate | sendemailvalidate |
+
+## Advice
+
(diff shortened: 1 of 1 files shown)


pepkaro:
If this sounds good to you, I've added a pull request with the given changes. Please let me know if there's anything I can update or change!

HOOKS.md
```diff
@@ -23,3 +23,16 @@ Husky supports all git hooks (https://git-scm.com/docs/githooks). Simply add the
...
+
+Advice on how to use specific hooks.
+
+### commit-msg
+
+Unlike the `commit-msg` git hook, Husky does not pass the commit message in to the `commitmsg` script. Scripts that want to examine the commit message must read the commit message itself. This can be done by reading `.git/COMMIT_EDITMSG`. For example, in Javascript, the commit message script could read the commit message with:
+
+    var vs = require('fs');
+
+    const commitMessage = fs.readFileSync('./git/COMMIT_EDITMSG', 'utf8');
```
jmlopez-rod: Should be `.git/COMMIT_EDITMSG'`.  as stated in above paragraph. And yes, this is good advice, if it wasn't for this I would still be trying to figure out how I can the file name since the git documentation says that it is passed as an argument.

Sufiane-Souissi: would be better to use `path.resolve` IMHO, I struggled a bit before using `path` ahah!

chanhosuh:
Is there a reason not to add this?  I found it helpful and I'm sure others would too, given that Git documentation says the hook takes in a command parameter.

HOOKS.md
```diff
@@ -23,3 +23,16 @@ Husky supports all git hooks (https://git-scm.com/docs/githooks). Simply add the
...
+
+## Advice
+
+Advice on how to use specific hooks.
+
+### commit-msg
+
+Unlike the `commit-msg` git hook, Husky does not pass the commit message in to the `commitmsg` script. Scripts that want to examine the commit message must read the commit message itself. This can be done by reading `.git/COMMIT_EDITMSG`. For example, in Javascript, the commit message script could read the commit message with:
+
+    var vs = require('fs');
```
TheBox193, line 35: Typo fix.
```suggestion
    const fs = require('fs');
```

typicode:
I think at the time I preferred to keep the docs small.
Thanks for the PR though! Closing it.

PR #194, opened 2017-10-09
Make initiating git command available inside hook

TxHawks:
Exports the git command that initiated the execution of the hook
available to the hook's code as `process.env.GIT_COMMAND`.

This can be accessed in regular git hooks through `ps -ocommand= -p $PPID`,
but is not available when running hook through an npm script.

It has quite a few use cases. For instance, when guarding against force
pushes in a `pre-push` hook, we can check if `process.env.GIT_COMMAND`
includes `--force ` or `-f`.

### DIFF
--- a/src/utils/get-hook-script.js
+++ b/src/utils/get-hook-script.js
@@ -94,6 +94,9 @@ module.exports = function getHookScript(hookName, relativePath, npmScriptName) {
       # Export Git hook params
       export GIT_PARAMS="$*"
 
+      # Export git command
+      export GIT_COMMAND=$(ps -ocommand= -p $PPID)
+
       # Run npm script
       echo "husky > npm run -s ${npmScriptName} (node \`node -v\`)"
       echo


dradetsky:
I came here looking for a way to guard against running postcheckout when `-b` is present. I approve of this.

TxHawks:
@typicode - if you'd like, I can open another pr agianst `dev` and add this change to `templates/hook.sh`

dradetsky:
@TxHawks it might be good to add a note to the details section of the readme, next to the bit about $GIT_PARAMS or something.

TxHawks:
@dradetsky - will do, as soon as @typicode indicates that he is interested in this

typicode:
@TxHawks Thanks for the PR.

Since `v0.15` is a complete rewrite already, for safety I prefer to fix potential bugs first before adding new features.

So this PR is probably going to be postponed until `v0.15` gets stable.

That said you mentioned use cases, besides `-f` what kind of other use cases do you see?

TxHawks:
@typicode  -
Sure thing, though it's really just a one-liner, and basically identical to exporting `$GIT_PARMS`, so hard to for me to imagine how it could introduce any bugs.

Other use cases, are, for instance, are checking for `checkout -b` like @dradetsky  suggested, checking for the deletion of permanent branches, forbidding or warning against `git commit -a` and probably many more.

TxHawks:
@typicode  - Also somewhat related - what do you think about my suggestion in [#166 (comment)](166#pullrequestreview-68390783)

jgornick:
As a workaround, in my bash script that husky calls, I use the following to get the `git push` command:

```shell
PUSH_PID=$PPID
PUSH_COMMAND=$(ps -ocommand= -p $PUSH_PID)

until [ "${PUSH_COMMAND/git push}" != "${PUSH_COMMAND}" ]; do
  PUSH_PID=$(ps -o ppid= -p $PUSH_PID)
  PUSH_COMMAND=$(ps -ocommand= -p $PUSH_PID)
done

echo "${PUSH_COMMAND}"
```

PR #204, opened 2017-11-12
Only install/uninstall hooks specified in config

xdissent:
Installing *all* the hooks by default can make things like rebasing super slow. This patch makes the installer only manage the hooks specified in the config.

### DIFF
--- a/src/installer/__tests__/index.ts
+++ b/src/installer/__tests__/index.ts
@@ -9,7 +9,9 @@ import { huskyIdentifier } from '../getScript'
 
 let tempDir
 
-const pkg = JSON.stringify({ husky: { skipCI: false } })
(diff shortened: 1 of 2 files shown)


AndersDJohnson:
Sometimes we install before we've setup any hooks, e.g., in a new project. I'm not sure this is the best default behavior, as users will have to re-install dependencies to get new hooks installed after adding.

tony:
@xdissent Can you `git pull --rebase origin master` (assuming origin is https://github.com/typicode/husky) / solve conflicts against the latest and force push to your branch?

tomasbjerre:
I rebased this commit on latest master, open a new PR with that: https://github.com/typicode/husky/pull/559

PR #209, opened 2017-11-23
Support more methods for defining config/hooks

vincecoppola:
Overview
-----
This PR makes it so that users can specify their hooks in a few more ways besides a top-level key in their `package.json`. This is accomplished by using [cosmiconfig](https://github.com/davidtheclark/cosmiconfig) which would add support for:
* A `.huskyrc`, `.huskyrc.json`, or `.huskyrc.y[a]ml` file
* A `husky.config.js` file that exports a config object

Motivation
-----
My team is already using husky and we want to move to the `beta` version due to its speed improvements but we're not too keen on adding a new top-level key in our `package.json`. We'd much prefer a `.huskyrc` file which would lead to a clearer separation of concerns and follow an existing pattern that is already in use by many packages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,8 @@ If you're upgrading from `0.14` simply move your hooks to `"husky"` field.
 }
 ```
 
+Alternatively, you may use any of the files/formats that are supported by [cosmiconfig](https://github.com/davidtheclark/cosmiconfig). This means that you can place your husky hooks config in a `.huskyrc` file or export them from a `husky.config.js` file as well. Cosmiconfig supports `js`, `json`, and `yaml`.
+
 ## Configure
 
 By default, husky expects your project's `package.json` and your `.git` directory to be at the same level. It can be configured to support monorepos or sub-directories.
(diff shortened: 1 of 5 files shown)


typicode:
Thank you!

typicode:
Just published new beta version `v0.15.0-beta.14` :)

vincecoppola:
Thank you for accepting the PR and for the quick turnaround @typicode!

3z3qu13l:
Hello,
Do you have an example of huskyrc.
@vinsidious our motivation are identical :)

typicode:
Hello @3z3qu13l,

Yes, here:
https://raw.githubusercontent.com/typicode/husky/master/.huskyrc

I'm using it to test on the CI, but basically the format is the same as in `package.json`. You just don't need to put `"husky"` field.

Here's another example using `.js`:

```js
// .huskyrc.js
module.exports = {
  hooks: {
    'pre-commit': 'npm test'
  }
}
```

PR #210, opened 2017-11-24
test: isCI should be defined

solcik:
Hello,

I was wondering why beta is trying to install on CI server (CircleCI) and failing due to hooks folder does not exist in the CI container. Problem is with the compilation from TS.

``` js
function install(gitDir, huskyDir) {
    console.log('husky > setting up git hooks');
    var userDir = pkgDir.sync(path.join(huskyDir, '..'));
    var conf = getConf(userDir);

    console.log(is_ci_1); // true
    console.log(is_ci_1.default); // undefined

    if (is_ci_1.default && conf.skipCI) {
        console.log('CI detected, skipping Git hooks installation"');
        return;
    }
```

### DIFF
--- a/src/installer/__tests__/index.ts
+++ b/src/installer/__tests__/index.ts
@@ -146,6 +146,13 @@ describe('install', () => {
   })
 })
 
+describe('isCI', () => {
+  it('should return boolean', () => {
+    expect(isCI).toBeDefined()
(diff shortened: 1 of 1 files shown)


typicode:
Good catch, thank you!

typicode:
Fix available in `0.15.0-beta.15` 👍

PR #214, opened 2017-12-08
fix(get-hook-script): use npm on PATH if it exists

jedwards1211:
instead of always loading and running nvm

### DIFF
--- a/src/utils/get-hook-script.js
+++ b/src/utils/get-hook-script.js
@@ -28,9 +28,12 @@ function platformSpecific() {
         # Node standard installation path /usr/local
         export PATH="$PATH:/usr/local/bin:/usr/local"
 
-        # Try to load nvm using path of standard installation
-        load_nvm ${home}/.nvm
-        run_nvm`
+        if ! which npm >/dev/null 2>&1
+        then
+          # Try to load nvm using path of standard installation
+          load_nvm ${home}/.nvm
+          run_nvm
+        fi`
     )
 
     return arr.join('\n')


jeremiebailleul:
Hi, 

I'm running on Windows at work and I don't have a choice regarding NodeJS installation directory (company configuration).

Therefore, when I perform a Git Push from IntelliJ, it fails to find NodeJS in "/c/Program Files/nodejs" and it skips the prepush git hook I've set up..

Could we have the possibility to also write NodeJS path in a .huskyrc for win32 platform ?
I will not be able to use Husky without this..

jedwards1211:
@typicode is there any problem with this or can it be merged?

rainboxx:
@jedwards1211 Could you tell more about your use case? How did you came to this change, what is the reason?

I wonder if the behavior after this PR is good or not. Personally I'd prefer to respect a `.nvmrc` even if there's `npm` in the path, since the file defines the version we want to run and test the source code. If the behavior is changed to never load `nvm` when `npm` is in the path, we'll have the same issues just the other way round. (As in "husky never loads the version specified in `.nvmrc`".)

jedwards1211:
@rainboxx actually I mistitled this issue, it's really about the node version that's on path.
Personally, I think all of your run and test scripts should check the node version themselves and error out if it's wrong -- otherwise when you're manually running the scripts you'll sometimes be confused if things go wrong with the node version is off (in my experience native modules are the primary source of confusion).

And actually, I think node and npm should check the `engines` field of the `package.json` and error out if it's wrong, like `yarn` does...

That said, you have a good point.  The primary reason I made this PR is really just that I'm tired of how long `nvm use` takes (and that `husky` unnecessarily calls it a second time, which is a separate issue).  It only takes a few seconds, but that few seconds feels longer and longer each time.

Perhaps this is a feature that needs to be customized, since not everyone will need to depend on husky to ensure the correct node version.

awilkins:
Slight segue : I really like `ndenv` for node version management, it doesn't have a noticeable time lag AFAICT and it doesn't require you to do an `nvm use` every time you switch folders, it "just works".

Trying out `nvm` was what got me associated with this thread of tickets because despite having removed `nvm` from my system, it's config files remained and husky was using them to choose which version of node to use, until I removed `~/.nvm`.

PR #217, opened 2017-12-25
Using Cross-platform unified shell.

gucong3000:


gucong3000:
Add support for bash script, just like:
```JSON
{
  "scripts": {
    "precommit": "eclint check $(git diff --cached --name-only)"
  }
}

```

PR #226, opened 2018-01-22
Update docs link in README

matchu:
Whoops, I clicked the docs link in the README, and it was a 404! Looks like the `docs.md` filename changed, so I updated it here, too.


typicode:
Thank you!

Issue #36, opened 2015-12-04
Don't pollute enclosing project hooks


ericsoco:
I just ran `npm install` on a package within my project's `node_modules`. This package has `husky` listed as a `devDependency`. `husky` ended up writing hooks into my project's `.git/hooks`, instead of the package within `node_modules`. Now I can't commit code in my own project without removing the hooks / `--no-verify`ing.

typicode:
Hi @ericsoco,

Not sure to understand why you want to run `npm install` in `node_modules`?
But if you're making some tests, you can run `npm rm husky --save-dev` to get rid of it and installed hooks.

ericsoco:
I was debugging a package and needed to rebuild it to pick up my changes.

My apologies, I don't know how husky does what it does, so I don't know if
it's feasible for it to be aware of context (installed within an npm
installed package or not), but it would be great if it could be smart
enough to not install hooks two levels up.

On Friday, December 4, 2015, typicode <notifications@github.com> wrote:

> Hi @ericsoco <https://github.com/ericsoco>,
>
> Not sure to understand why you want to run npm install in node_modules?
> But if you're making some tests, you can run npm rm husky --save-dev to
> get rid of it and installed hooks.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/typicode/husky/issues/36#issuecomment-161933831>.
>


-- 
ditditdit on my mobile phone

iamstarkov:
cant wrap my head about this problem. @ericsoco can you elaborate a bit?

ericsoco:
It's been a long time @iamstarkov, but best as I remember, my initial comment describes it well -- I `npm install`ed a package in my project that uses `husky`. In order to debug that package, I had to run `npm install` from within its folder (e.g. `myproject/node_modules/package-that-uses-husky`). Doing so installed git hooks on my project (`myproject/.git/hooks`) instead of on the package (`myproject/node_modules/package-that-uses-husky/.git/hooks`). That made things go wonky on my project.

If possible, `husky` should know where to install its hooks to avoid this problem. This seems an edge case, though.

ljharb:
I had the same problem (with `ghooks`). This caused git hooks I didn't write, audit, or know about to be installed in my company's private repo.

Automatically injecting git hooks on install (as opposed to requiring the package consumer to explicitly add a command in a postinstall script, which would be far safer) is dangerous enough - it should certainly not traverse above the folder the closest `package.json` (that references husky) is in under any circumstances.

[...]

PR #244, opened 2018-02-27
Allow installation in the outer .git directory

TETRA2000:
Hi! I'm participated in a project which has separated frontend/backend directories.
It has a frontend directory as a subdirectory. And I would like to install hooks from the subdirectory.

```
└───some_backend_repo
    └───.git
    └───frontend
        └───package.json
        └───node_modules
        └─── ....
```

But husky doesn't allow me to install hooks when `package.json` is located in another place of `.git` located.

So I want an opted-in option to install hooks in the outer .git directory.

I implemented to use env variables. So I think this doesn't make bad effects when unwanted.(like #36 )

I hope you to review this pull request.
Thank you.

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -16,6 +16,14 @@ If you don't want `husky` to automatically install Git hooks, simply set `HUSKY_
 HUSKY_SKIP_INSTALL=true npm install
 ```
 
+## Allow installation in the outer .git directory
+
+If you want to install hooks in outer git repository, set `HUSKY_ALLOW_INSTALL_OUTER` environment variable to `true`.
+
+```
+HUSKY_ALLOW_INSTALL_OUTER=true npm install
+```
+
 ## Multi-package repository (monorepo)
 
 If you have a multi-package repository:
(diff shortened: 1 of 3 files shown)


typicode:
Hi @TETRA2000,

Thank you for the PR, have you seen https://github.com/typicode/husky/blob/dev/docs.md#sub-directory-package?
Would that work for you?

TETRA2000:
@typicode Thank you for replying.
It worked for me. But I  don't want to add `package.json` in root directory in this case. 
Because root directory contains files which only used by back-end application except `frontend` directory.
And I'm thinking to trigger front-end related tasks(`npm run lint` and so on) by git hook.

Also I know this kind of specification may cause unintended behavior.
So, could you give me your opinion?

Thank you.

malixsys:
@typicode @TETRA2000 
You should use `git rev-parse --show-toplevel` from a child process to find where the git root is...

typicode:
Fixed by `1.0`

Issue #93, opened 2017-01-10
Expose git stdin parameters for pre-push


Necroskillz:
As you can see in the pre-push sample: https://github.com/git/git/blob/master/templates/hooks--pre-push.sample, they use some stdin parameters to determine local and remote sha and ref.

Main usecases for having these parameters are:
 - To not execute pre-push hook when you are deleting a remote branch (this is the most important, since it's kind of annoying when hooks starts when you just want to delete a branch)
 - To filter branch names which you want to execute pre-push hooks on

callmephilip:
+1: I am trying to determine the name of the branch in pre-push and cannot seem to find a way to do this with husky ($GIT_PARAMS only contains information regarding remote + repository uri)

okonet:
I have similar use-case with post-checkout: https://gist.github.com/okonet/9706ec3b7f29e05109f3744fbc1fc91e

So I think parameters should be exposed for all hooks by default.

ghooks package does it but it's deprecated in favor of this one.

PR #253, opened 2018-03-13
add GIT_STDIN env variable

testerez:
From git's [pre-push hook documentation](https://git-scm.com/docs/githooks#_pre_push
):
> Information about what is to be pushed is provided on the hook’s standard input with lines of the form: `<local ref> SP <local sha1> SP <remote ref> SP <remote sha1> LF`

With this PR, stdin is stored in GIT_STDIN env variable so user hooks can access it.

Should fix: https://github.com/typicode/husky/issues/93

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -4,9 +4,9 @@
 
 `husky` supports all Git hooks defined [here](https://git-scm.com/docs/githooks).
 
-## Access Git params
+## Access Git params and stdin
 
-You can access them using `GIT_PARAMS` environment variable.
+You can access them using `GIT_PARAMS` and `GIT_STDIN` environment variables.
 
 ## Disable auto-install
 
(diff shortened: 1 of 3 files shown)


typicode:
Thank you! :)

simonplend:
@typicode I've just hit this same issue when trying to write a `pre-push` hook. This change by @testerez looks good, but I noticed that since merging this PR you've reverted the changes: https://github.com/typicode/husky/commit/da7a9e3914705a6c93080c4fc3d1cb636cbb994e#diff-cb1b5fb90009466f49e307c8430feb5e

Is this going to come back? I'm a little bit stuck right now, but I'd really like to keep using husky!

testerez:
@typicode did you notice a bug with my code?

turadg:
It's `HUSKY_GIT_STDIN` as of https://github.com/typicode/husky/blob/dev/CHANGELOG.md#100-rc8

Bkucera:
@turadg @testerez there hasn't been a release since this was merged: https://github.com/testerez/husky/releases

PR #257, opened 2018-03-22
Update dependent repositories number at readme

KnupMan:
According to https://libraries.io/npm/husky/dependent-repositories, `husky` is used by more than 18k repositories
image


typicode:
> Unbelievable!

Yes, that's crazy 😄 really happy that people appreciate this project ❤️

PR #266, opened 2018-04-07
Fix: fatal: Not a git repository: '.git'

huixisheng:
see [https://stackoverflow.com/questions/4043609/getting-fatal-not-a-git-repository-when-using-post-update-hook-to-execut](https://stackoverflow.com/questions/4043609/getting-fatal-not-a-git-repository-when-using-post-update-hook-to-execut)


RickBross:
@typicode  @huixisheng  is this a viable fix?

I need to deploy a React app ASAP...

Thanks!

typicode:
Not sure what would be the side effect of this and v4 is in maintenance mode.
Thanks for the PR.

PR #294, opened 2018-05-19
Prevent package normalizing

TrySound:
Currently date-fns has this `package.json` and husky fails because the version is invalid.

```js
{
  "version": "DON'T CHANGE; IT'S SET AUTOMATICALLY DURING DEPLOYMENT; ALSO, USE YARN FOR DEVELOPMENT"
}
```

I think that husky shouldn't be cared about package linting. Husky is used by user for any kind of linting and shouldn't force user to anything.


typicode:
You're right, thank you!

PR #300, opened 2018-06-08
Removes console.log

philals:
Hi

Thanks for making husky.

I found this log to create noise on my terminal when using a `pre-commit` hook.

Thoughts on removing it? Or what about a `silentLogs: true` config option? Happy to code it.

### DIFF
--- a/src/runner/index.ts
+++ b/src/runner/index.ts
@@ -43,7 +43,6 @@ export default async function(
     }
 
     if (command) {
-      console.log(`husky > ${hookName} (node ${process.version})`)
       execa.shellSync(command, { cwd, env, stdio: 'inherit' })
       return 0
     }


philihp:
Yes please! Silent by default. http://www.linfo.org/rule_of_silence.html

knod:
A `debug` or `verbose` config item sounds more flexible. That way it can be used for all logs.

AlaaAttya:
+1 for the verbose flag

Swiftwork:
I find this troublesome as well. It seems as if this logging interferes with error reporting in VSCode. Instead of receiving a usable error message the user is presented with this pop up.

Screenshot 2019-09-10 at 23 33 32

It would seem to me that VSCode displays the first log it receives which will always be the debug message rather than the actual written error message.

/Swiftwork

IanEdington:
Is there any movement on this PR? Husky is currently breaking other command line tools because of not being silent. This issue has come up by a number of people since 2016. Why has there not been movement on a fix?

If there's a plan on how to solve this I have time to work on a fix. Otherwise what do we need to make a decision?

typicode:
Husky 5 is silent by default, thanks for the PR and feedback :+1:

PR #343, opened 2018-09-07
Adds "exec < /dev/tty" for hooks that require user input

joelgallant:
See: https://stackoverflow.com/a/3533735

I had a git hook that prompts the user, and it fails early. Running it in any shell otherwise worked fine.


joelgallant:
Unit tests on my machine failed (something to do with run-node), I tried updating the snapshot that should be affected manually.

joelgallant:
Ah, seems like CI is broken right now

yvele:
Is that integrated to the new release? I can't see any mention of it in the changelog https://github.com/typicode/husky/blob/master/CHANGELOG.md 🤔

typicode:
My bad, didn't really want to close the issue, but deleting the `dev`  has automatically done it (which is logical ^^').

The issue with this is that while it works in a shell, I'm not sure it won't cause issue in a GUI Git client (e.g. GitHub Desktop). Since you can't type text, I think it'll just block the commit.

joelgallant:
Could this be changed to `[ -t 0 ] && exec < /dev/tty` to avoid happening in non-tty environments?

typicode:
Worth testing :+1:
What do you use to prompt the user? A shell script or npm module? It's to create a minimal test.

joelgallant:
Personally I use an npm module with `inquirer`, I can try to add a test in to this for TTY, maybe using `spawn` options to emulate both environments

typicode:
For the moment, I think first test would be to validate that adding `[ -t 0 ] && exec < /dev/tty` doesn't break anything with GitHub client. I mean, no need to spend time/energy in some automated tests if the first step doesn't work.

joelgallant:
Just rebased on `master`. I tried the `[ -t 0 ]` approach, but it turns out that git hooks are not run in tty by design, `-t 0` is false.

So the alternative seemed like adding a config option, for users to opt-in to the functionality. I don't know this code base, but made my best attempt at following the flow and conventions.

joelgallant:
Looks like I didn't notice https://github.com/typicode/husky/pull/166 before, which is essentially the same thing

typicode:
See https://github.com/typicode/husky/pull/415

PR #344, opened 2018-09-09
advise '--no-verify' does not skip post-* hooks

### LINKED ISSUE #342: change bypass warning: git post-* hooks do not respect `--no-verify`
### problem
Husky 0.14.x and master warns `(add --no-verify to bypass)` on all hooks except prepare-commit-msg.  No post-* hook respects --no-verify, so this message should be omitted.

### Evidence
- https://git-scm.com/docs/githooks does not mention `--no-verify` in any post-* hook.

- Test:
```{sh}
#!/bin/sh
# setup
mkdir husky-test && cd husky-test;
git init
# create initial commit to play around in
function useless_commit () {
  echo "$1" >> foo && git add foo && git commit -m "$1"
}
useless_commit "initial commit"

# husky setup
cat << 'eof' > ./package.json
{
  "name": "husky-test",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "postcommit": "exit 1",
    "postcheckout": "exit 1",
  },
  "keywords": [],
  "author": "",
  "license": "ISC",
  "devDependencies": {
    "husky": "^0.14.3"
  }
}
eof
npm install

useless_commit "foo" # fails, but still warns about no-verify
git checkout HEAD^ #   fails, but still warns about no-verify
# same for post-merge, post-push
``` 
### solution
The code to change is at [src/runner/index.ts#L72](https://github.com/typicode/husky/blob/495f4a0ee80118d2ac146dd8e1ae62b9c8fd4f18/src/runner/index.ts#L72).  The issue which prompted that change is #137. I'll see if I can get to fixing this later tonight.

SKalt:
Closes #342.

### DIFF
--- a/src/runner/index.ts
+++ b/src/runner/index.ts
@@ -68,10 +68,14 @@ export default async function run(
 
     return 0
   } catch (err) {
-    const noVerifyMessage =
-      hookName === 'prepare-commit-msg' || hookName.includes('post')
-        ? '(cannot be bypassed with --no-verify due to Git specs)'
-        : '(add --no-verify to bypass)'
+    const noVerifyMessage = [
+      'commit-msg',
+      'pre-commit',
+      'pre-rebase',
+      'pre-push'
+    ].includes(hookName)
+      ? '(add --no-verify to bypass)'
+      : '(cannot be bypassed with --no-verify due to Git specs)'
 
     console.log(`husky > ${hookName} hook failed ${noVerifyMessage}`)
     return err.code
--- a/src/runner/index.ts
+++ b/src/runner/index.ts
@@ -69,7 +69,7 @@ export default async function run(
     return 0
   } catch (err) {
     const noVerifyMessage =
-      hookName === 'prepare-commit-msg'
+      hookName === 'prepare-commit-msg' || hookName.includes('post')
         ? '(cannot be bypassed with --no-verify due to Git specs)'
         : '(add --no-verify to bypass)'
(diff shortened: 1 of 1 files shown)


typicode:
Hi @SKalt,

Really good catch and thanks for PR 👍 

After checking each Git command referenced in https://git-scm.com/docs/githooks, and except if I've missed something, it seems like only these hooks can be bypassed.

https://git-scm.com/docs/git-commit#git-commit---no-verify

> This option bypasses the __`pre-commit`__ and __`commit-msg`__ hooks

https://git-scm.com/docs/git-rebase#git-rebase---no-verify

> This option bypasses the __`pre-rebase`__ hook

https://git-scm.com/docs/git-push#git-push---no-verify

> Toggle the __`pre-push`__ hook

Could you modify the PR so that it displays `(add --no-verify to bypass)` only if it's one these 4 hooks?

SKalt:
Sure thing! Would you accept checking an environment variable
HUSKY_NO_VERIFY or GIT_NO_VERIFY to bypass the other hooks as a part of
this issue, or should I raise a separate issue for discussion?

typicode:
I'd rather separate as this one fixes an issue/bug and will be merged in any case, and the other needs consideration.

SKalt:
Fixes added and a separate issue raise in #346 .

typicode:
Awesome, thanks!

PR #353, opened 2018-09-24
Add info about husky-upgrade with npx to README

vikr01:
In certain scenarios `node_modules` might not be in the current directory (i.e. `yarn` workspaces, user's cwd is not the root of the package, or `husky` installed somewhere higher up).

`npx` will work so long as the user is able to connect to the `npm` registry, so it's good to note how to do `husky-upgrade` through it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,8 +64,14 @@ Simply move your existing hooks to `husky.hooks` field and use raw Git hooks nam
 
 Alternatively, you can run the following command which will do the same automatically for you ;)
 
-```
+```sh
 ./node_modules/.bin/husky-upgrade
+```
+
+Or with `npx`:
+
+```sh
+npx husky husky-upgrade
 ```
 
 Starting with `1.0.0`, you can also configure hooks using `.huskyrc`, `.huskyrc.json` or `.huskyrc.js` file.


typicode:
Hi @vikr01,

I haven't thought about it, thanks :+1:. I've tried with just `npx husky-upgrade` after installing `husky` and it works, would that work too in the case you describe?

My only concerns with `npx husky husky-upgrade` is that you can run `husky-upgrade` in a repository that doesn't have `husky` installed or maybe run a different version of `husky-upgrade` than the one installed.

vikr01:
@typicode Yes, this is a solution for people who don't have the `node_modules` in their current directory.

Unfortunately, yes it would work on repos without `husky` and ones with a different version of `husky`. 

However, this could be checked by doing something like:

```javascript
const {version: thisHuskyVersion} = require('../package.json');
const {version: huskyVersion} = require('husky/package.json');
if(thisHuskyVersion !== huskyVersion) {
  throw new Error(
    `This version of husky is not compatible with your current version, use npx husky@${
      huskyVersion
    } husky-upgrade instead.`
  );
}
```

Another possible solution is to use:

```sh
$(node -p "require.resolve('husky/lib/upgrader/bin.js')")
```

Though I'm not sure about the compatibility of the above.

PR #359, opened 2018-09-26
perf(husky-upgrade): don't create husky config if no hooks exist

sibiraj-s:
Don't create husky config in package.json if hooks doesn't exist already


typicode:
Thanks 👍

PR #362, opened 2018-09-26
upgrader: Add trailing newline to package.json

JHilker:
By default npm includes a trailing newline when it writes the `package.json`. this preserves that behavior to not cause churn in the file in the most common case.

https://github.com/npm/cli/blob/4c65cd952bc8627811735bea76b9b110cc4fc80e/lib/install/update-package-json.js#L48


typicode:
Thank you! 👍

PR #292, opened 2018-05-18
Fix: Create .git/hooks if doesn't exist

luiscoms:

PR #366, opened 2018-09-29
Do not fail just because "hooks" foder does no exists yet

DavidBabel:
The way we all use husky is by "npm install".

It is very annoying when for a reason, the hooks does not install. Then we have to go to every developers to ask them to check their install. Which is very not convenient.

This proposal, is to remove one fail, which is, in my opinion, not so difficult to handle.
Is ".git" exists, but ".git/hooks" does not, instead of skipping installation, we create the folder. Not a big deal, but sooooo annoying the current way.

### DIFF
--- a/src/installer/index.ts
+++ b/src/installer/index.ts
@@ -174,9 +174,7 @@ export function install(
   }
 
   if (!fs.existsSync(path.join(resolvedGitDir, 'hooks'))) {
-    console.log(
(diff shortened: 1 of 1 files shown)


typicode:
Thanks for the PR, going to close as there was a similar PR (https://github.com/typicode/husky/pull/292)

DavidBabel:
nice no prob

PR #368, opened 2018-10-01
Create .git/hooks if doesn't exist

luiscoms:
Recreating #292


typicode:
Thanks a lot! And sorry for the mistake :)

PR #373, opened 2018-10-05
don't default to run-node, use 'node' if it works

ivanov:
If `node` works, we will not bother with `run-node` on non-windows
platforms, since `run-node` requires `bash` and operating systems like
OpenBSD and FreeBSD, as well as minimal docker images like Alpine,
don't have bash in their base installations.

### DIFF
--- a/src/installer/__tests__/__snapshots__/getScript.ts.snap
+++ b/src/installer/__tests__/__snapshots__/getScript.ts.snap
@@ -8,13 +8,16 @@ exports[`hookScript should match snapshot (OS X/Linux) 1`] = `
 scriptPath=\\"node_modules/husky/run.js\\"
 hookName=\`basename \\"$0\\"\`
 gitParams=\\"$*\\"
+node=node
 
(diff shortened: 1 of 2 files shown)


ivanov:
I opened this PR up to start a conversation. I am unable to use a project which uses husky in a Unix/Linux setting that does not have bash (OpenBSD, FreeBSD, and Alpine, in my case). Husky already tries to be a good POSIX citizen by using `sh` as the shell for its script, but unfortunately upstream `run-node` is not so generous and requires `bash`. This PR is one way of mitigating that - just using `node` if that works (and therefore bypassing `run-node`'s usage of `bash`.

ivanov:
My machine was confused about timezones while I made this, so the last commit ended up with a timestamp hours before the first two. I've squashed them into one commit now.

typicode:
Interesting use case. Husky 5, being less opinionated and simpler, should be exempt from this problem so closing it.
Thanks for the PR :+1:

PR #377, opened 2018-10-08
Print debug message at hook start

### LINKED ISSUE #371: When HUSKY_DEBUG is true, echo message at the beginning of the shell script
This should help troubleshoot some issues by confirming that corresponding Git hook is started.

```sh
$ HUSKY_DEBUG=true git commit -m '...'
husky:debug pre-commit
```

The file to be modified:
https://github.com/typicode/husky/blob/master/src/installer/getScript.ts#L24

wsmd:
Closes #371

This PR prints added ability to echo a message at the beginning of the shell script when `HUSKY_DEBUG` is set.

**Example**

```
$ HUSKY_DEBUG=true git commit -m 'accidental commit'

husky:debug pre-commit hook started...

husky > pre-commit (node v10.11.0)

> husky@1.1.1 test /Users/waseem/Repositories/husky
> npm run lint && jest


> husky@1.1.1 lint /Users/waseem/Repositories/husky
> tslint 'src/**/*.ts'

...
```

### DIFF
--- a/src/installer/getScript.ts
+++ b/src/installer/getScript.ts
@@ -21,6 +21,11 @@ ${huskyIdentifier}
 scriptPath="${script}.js"
 hookName=\`basename "$0"\`
 gitParams="$*"
+
+if [[ ! -z "$\{HUSKY_DEBUG\}" ]]; then
+  echo "\nhusky:debug $hookName hook started...\n"
+fi
+
 ${
   platform !== 'win32'
     ? `
--- a/src/installer/__tests__/__snapshots__/getScript.ts.snap
+++ b/src/installer/__tests__/__snapshots__/getScript.ts.snap
@@ -9,6 +9,13 @@ scriptPath=\\"node_modules/husky/run.js\\"
 hookName=\`basename \\"$0\\"\`
 gitParams=\\"$*\\"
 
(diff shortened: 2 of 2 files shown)


src/installer/getScript.ts
```diff
@@ -21,6 +21,11 @@ ${huskyIdentifier}
 scriptPath="${script}.js"
 hookName=\`basename "$0"\`
 gitParams="$*"
+
+if [[ ! -z "$\{HUSKY_DEBUG\}" ]]; then
```
wsmd: This will only check if the environment variable does not have a zero length, not if it's set to `"true"`.

```sh
# all of these will result in printing a debug message
HUSKY_DEBUG=1
HUSKY_DEBUG=0
HUSKY_DEBUG=asdf1234teralyanything
HUSKY_DEBUG=true
HUSKY_DEBUG=false # this will still show a debug message!
```

The current implementation is somewhat similar to this:

https://github.com/typicode/husky/blob/b7d4a69f6d0e852b046e1ea21501c333cecf7630/src/installer/bin.ts#L6-L8

Is there a need to check if the variable is explicitly set to `true`?

typicode: I think it's fine like this 👍

typicode:
Thanks a lot for the PR. I'll merge it later.

Issue #380, opened 2018-10-18
PR available: add HUSKY_GIT_HOOK environment variable so multi-hook scripts can be used again


bartvanandel:
We are using one script for multiple git hooks. Previously we were able to distinguish which hook was being called by looking at `process.env.npm_lifecycle_event`, however since Husky 1.0 doesn't use the script section anymore, this value is no longer present. Husky for sure knows which hook is being called, I suggest this to be added to the environment as `HUSKY_GIT_HOOK` so we can again use multi-purpose hook scripts.

bartvanandel:
@typicode I'm hoping you can look into this within a short time frame. We had to upgrade Husky for a different reason but now our hooks are actually broken :(

bartvanandel:
It would be nice if the maintainer could take a look at this?

bartvanandel:
We are currently using something like the latter. The former isn't portable across platforms without something like crossenv. Having this built-in would be nice but at this moment we can live without it.

Thanks for your reply.

typicode:
Cleaning old issues. 

That said, it should be possible to have a generic way to get hook name with v5 using `basename "$0"`:

```
echo "hook name is $(basename "$0")"
```

PR #381, opened 2018-10-18
feat: store hook name in HUSKY_GIT_HOOK environment variable

bartvanandel:
Add `HUSKY_GIT_HOOK` environment variable so hook scripts can easily see which hook is being executed.

Fixes https://github.com/typicode/husky/issues/380

### DIFF
--- a/DOCS.md
+++ b/DOCS.md
@@ -2,7 +2,7 @@
 
 ## Supported hooks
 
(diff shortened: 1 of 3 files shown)


bartvanandel:
@typicode Are you monitoring PRs?

PR #386, opened 2018-10-26
pipe /dev/tty into commit hook command

### LINKED ISSUE #385: process.stdin in the `commit-msg` hook script is not a tty
I am using [commitizen](https://github.com/commitizen/cz-cli) to prompt user input during commit-msg hook, however the `process.stdin` of the process is not tty, ie, i couldn't get user keyboard input from the process.

tanhauhau:
Fixes #385

### DIFF
--- a/src/installer/__tests__/__snapshots__/getScript.ts.snap
+++ b/src/installer/__tests__/__snapshots__/getScript.ts.snap
@@ -18,7 +18,7 @@ if ! command -v node >/dev/null 2>&1; then
 fi
 
 if [ -f $scriptPath ]; then
-  node_modules/run-node/run-node $scriptPath $hookName \\"$gitParams\\"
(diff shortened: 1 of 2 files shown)


typicode:
Thanks for the PR 👍

While it works in terminal, I think it will fail in GUI clients. Quickly tested with VSCode Git interface and got `cannot open /dev/tty: No such device or address`.

I don't know for a way to test if `tty` is available before using it.

But to allow people to enable it locally, I'm experimenting with sourcing a `~/.huskyrc` and let people add `exec < /dev/tty` if they need it.

ikokostya:
> I don't know for a way to test if tty is available before using it.

It can be tested using `test -t` command in `shell`: https://stackoverflow.com/a/911213

typicode:
Thanks @ikokostya. Re-opening it may need more testing actually.

> I think that executable program should check it itself.

Which one?

typicode:
Will be added in https://github.com/typicode/husky/pull/415

Issue #330, opened 2018-08-17
Running multiple command on a hook


Quadriphobs1:
I want to run more than one command on a hook. For example running `npm run lint and npm run test` in a `pre-push`.

Keramet:
in package.json:
```
...
"scripts": {
    "precommit": "npm test && npm run lint",
    "start": ...,
    "lint": ... 
  },
```

Quadriphobs1:
Ohhh I was actually not using the `&&` for it and I hope it would work way

djErock:
How can I run these (npm run lint && npm test) concurrently?

Thomaash:
Try `run-p lint test` from https://www.npmjs.com/package/npm-run-all.

Issue #319, opened 2018-07-17
Support for multiple scripts per hook?


cviebrock:
I was hoping something like this was supported:

```json
{
  "husky": {
    "hooks": {
      "commit-msg": [
        "some-script.sh",
        "some-other-script.sh",
        "one-more-script.sh",
      ]
    }
  }
}
```

And Husky would run those 3 scripts in order, allowing git to stop at the first failure.

I realize this _could_ be done with one script (that itself runs all 3 parts), or by doing something like:

```json
"commit-msg": "some-script.sh && some-other-script.sh && one-more-script.sh"
```

But I thought the json array might be a bit cleaner syntax.

Any thought to implementing something like this?

typicode:
Hi @cviebrock,

No plan to support this for the moment in Husky.

Maybe there are existing tools or task runners that would allow defining scripts this way and could be used with Husky.

cviebrock:
Thanks for the answers. I might try my hand at a PR to add this, but in the meantime https://github.com/okonet/lint-staged seems to do what I need.

dianasaurio:
I think this would be a nice feature for husky.

typicode:
> Thanks for the answers. I might try my hand at a PR to add this, but in the meantime https://github.com/okonet/lint-staged seems to do what I need.

Glad to read it :)

> I think this would be a nice feature for husky.

Realized recently that you can do it somehow with a `.huskyrc.js` file:

```js
const tasks = t => t.join(' && ') 

module.exports = {
  hooks: {
    'pre-commit': tasks([
      'echo 1',
      'echo 2',
      'echo 3'
    ])
  }
}
```

kalpeshsingh:
> > Thanks for the answers. I might try my hand at a PR to add this, but in the meantime https://github.com/okonet/lint-staged seems to do what I need.
> 
> Glad to read it :)
> 
> > I think this would be a nice feature for husky.
> 
> Realized recently that you can do it somehow with a `.huskyrc.js` file:
> 
> ```js
> const tasks = t => t.join(' && ') 
> 
> module.exports = {
>   hooks: {
>     'pre-commit': tasks([
>       'echo 1',
>       'echo 2',
>       'echo 3'
>     ])
>   }
> }
> ```

For `.huskyrc` file - 

```
"hooks": {
    "pre-commit": "npm run format && npm run test"
}
```

PR #406, opened 2018-11-24
feat(runner): support hooks with multiple commands

eranshmil:
Added support to run hooks with multiple commands, except for old commands.

Related issues: #330 #319

### DIFF
--- a/src/runner/index.ts
+++ b/src/runner/index.ts
@@ -29,13 +29,13 @@ export default async function run(
   }
   const config = getConf(cwd)
 
-  const command: string | undefined =
+  let commands: string | string[] | undefined =
     config && config.hooks && config.hooks[hookName]
 
   const oldCommand: string | undefined =
     pkg && pkg.scripts && pkg.scripts[hookName.replace('-', '')]
 
-  // Run command
+  // Run commands
   try {
     const env: IEnv = {}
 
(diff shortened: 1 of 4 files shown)


nunofgs:
Need this! When is it going to be merged? 🙃

joseSantacruz:
+1

src/runner/index.ts
```diff
@@ -51,9 +51,18 @@ export default async function run(
...
+    if (commands) {
+      if (!Array.isArray(commands)) {
+        commands = [commands]
+      }
+
+      commands.forEach((command: string) => {
+        console.log(
+          `husky > ${hookName} > ${command} (node ${process.version})`
+        )
+        execa.shellSync(command, { cwd, env, stdio: 'inherit' })
```
datashaman: What if a command fails?

eranshmil: Yeah, you're right.
I'll change that if the PR will be relevant.

G-Rath:
Love this! Anything I can do to help get it across the line? (I was all set to fork to implement this before I saw this PR 😂)

Also, I'm not sure who's managing it, but the [`huskyrc` schema](http://json.schemastore.org/huskyrc) should be updated before this is released :D

pke:
Is the missing schema update blocking this from being merged?

PR #423, opened 2018-12-13
Fix whitespace in project folder name issue

klmdb:


typicode:
Thanks for the PR 👍

Hmm don't really know why it fails with snapshots, they seem good in the PR.

Usually, you can leave the package version as it is. This way project authors can bump the version just when needed.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "husky",
-  "version": "1.2.0",
+  "version": "1.2.1",
```
typicode:
```suggestion
  "version": "1.2.0",
```

typicode:
Can you try running `npm test -- -u` or `npx jest -u` to update snapshots, it should fix the failing CI.

src/installer/__tests__/__snapshots__/getScript.ts.snap
```diff
@@ -28,12 +28,12 @@ if ! command -v node >/dev/null 2>&1; then
...
 fi
 
-if [ -f $scriptPath ]; then
+if [ -f "$scriptPath" ]; then
   if [ -f ~/.huskyrc ]; then
     debug \\"source ~/.huskyrc\\"
     source ~/.huskyrc
   fi
-  node_modules/run-node/run-node $scriptPath $hookName \\"$gitParams\\"
+  node_modules/run-node/run-node "$scriptPath" $hookName \\"$gitParams\\"
```
typicode: Just realized that it was failing because `"` isn't escaped, but adding `-u` to `npm test` should update them correctly :)

klmdb:
`npm test -- -u` did the trick

typicode:
Thanks :)

PR #433, opened 2018-12-22
Make `HUSKY_DEBUG=1` effective like `HUSKY_DEBUG=true`

Calinou:
As part of a troubleshooting routine, I set `HUSKY_DEBUG` to 1 and wondered why it didn't have any effect. It turns out Husky only enables debugging if `HUSKY_DEBUG` is set to `true`, which is inconsistent with most UNIX utilities I know of (where `1` and `true` are both seen as "truthy" values).


typicode:
Thanks for the PR

@Calinou out of curiosity, do most of the UNIX utilities also support `true` or just `1`?

Calinou:
> Which UNIX utilities are you using?

I don't have a whole lot of specific examples, but for instance, true color can be enabled in the [micro](https://micro-editor.github.io/) editor by setting `MICRO_TRUECOLOR` to 1.

PR #451, opened 2019-02-01
replace please-upgrade to forked one

a-x-:
original please-upgrade is incorrect due to cannot work with semver complient expressions


a-x-:
looks like please-upgrade added semver, but had not updated his readme
https://github.com/typicode/please-upgrade-node/issues/9

PR #455, opened 2019-02-09
Add configuration for choosing the branches for which the hook commands should be executed

mikailcf:
## Description
Adding a simple configuration field for when you don't want to run the configured hooks for every branch.

There's a check verifying if the hook commands should be executed for the current working branch.

## How to use
Add the key `active-branches` and set an array value with the name of the branches for which you want the hooks executed.

Example:
```
// package.json
{
  "husky": {
    "hooks": {
      "pre-commit": "npm test",
      "pre-push": "npm test",
      "active-branches": ["master", "staging"],
      "...": "..."
    }
  }
}
```

### DIFF
--- a/src/runner/index.ts
+++ b/src/runner/index.ts
@@ -35,6 +35,17 @@ export default async function run(
   const oldCommand: string | undefined =
     pkg && pkg.scripts && pkg.scripts[hookName.replace('-', '')]
 
+  const targetBranch: string | undefined =
+    config && config.hooks && config.hooks['target-branch']
+
+  const currentBranch = execa.shellSync(
+    'env -i git rev-parse --abbrev-ref HEAD'
+  ).stdout
+
+  if (targetBranch && targetBranch !== currentBranch) {
+    return 0
+  }
+
   // Run command
   try {
     const env: IEnv = {}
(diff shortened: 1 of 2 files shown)


src/runner/index.ts
```diff
@@ -35,6 +35,17 @@ export default async function run(
   const oldCommand: string | undefined =
     pkg && pkg.scripts && pkg.scripts[hookName.replace('-', '')]
 
+  const activeBranches: string | undefined =
```
pamepeixinho: Maybe it can be an array of strings. What do you think?

mikailcf: It should indeed be an array, that code would result in an error. Thanks for pointing it out!

beppe9000:
Is it possible instead to expand this in order to specify different scripts for different branches ?

e.g.

```
// package.json
{
  "husky": {
    "hooks": {
      "pre-commit": "npm test",
      "pre-push": "npm test",
      "active-branches": ["master", "staging"],
      "branch-scripts": {
          "master": {
                           "pre-commit": "ng lint",
                           "pre-push": "npm test"
                         }
          "staging": {
                           "pre-commit": "ng lint",
                           "pre-push": "npm xyz"
                         }
       }
      "...": "..."
    }
  }
}
```

muntasirsyed:
Love this! I do agree with @beppe9000 that configuration per branch would perhaps be slightly more useful for most use cases.

beppe9000:
is this implemented ?

PR #490, opened 2019-05-16
Allow husky configs from child package.json

DanElliottPalmer:
## Description

The team I work with is currently having issues with Husky in a monorepo like structure. We have a parent project with a set of Husky definitions in its package.json. We also have a child related package in another folder with its own dependencies and scripts in the same repo. We would like to have different set of Husky hooks for the child package.

## Changes

Instead of Husky using the directory of where the .git is located, this will take into account the cwd of the user where they triggered the git hook from. It will then look for the closest package.json from that point.

## A bit more detail

When a git hook runs, it changes the working directory:

> Before Git invokes a hook, it changes its working directory to either $GIT_DIR in a bare repository or the root of the working tree in a non-bare repository. An exception are hooks triggered during a push (pre-receive, update, post-receive, post-update, push-to-checkout) which are always executed in $GIT_DIR
> https://git-scm.com/docs/githooks

To help identify the location of the user, $GIT_PREFIX is set specifying the original path relative to .git (https://github.com/git/git/blob/6f9504c48ea59951a2aa3ee17e7c7fc36285e225/Documentation/config.txt#L762-L763). We can use this information to work out which package.json we want to be reading the config from.

## Outstanding questions/concerns
1. This is quite a big behaviour change.
2. If it finds a child package.json that contains no husky config, should it continue moving up the folder tree till it finds a package.json with a husky config? Obviously stopping at the .git folder level.
3. I have not had a chance to test this on Windows and wondered if you have a preferred method for testing?

### DIFF
--- a/src/installer/__tests__/__snapshots__/getScript.ts.snap
+++ b/src/installer/__tests__/__snapshots__/getScript.ts.snap
@@ -27,6 +27,16 @@ if ! command -v node >/dev/null 2>&1; then
   echo \\"Info: Can't find node in PATH, trying to find a node binary on your system\\"
 fi
 
+
+# Git hooks change the PWD to be the dir where the .git folder lives. GIT_PREFIX
+# stores the original location of where the hook was executed from. This allows
+# us to work out the original PWD.
+gitCwd=\\"$GIT_PREFIX\\"
+if [ ! -z \\"$gitCwd\\" ] then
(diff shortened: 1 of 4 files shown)


hilarykitz:
+1 this would be very useful

DanElliottPalmer:
Closing this PR as its been almost year and doesn't seem like there's any appetite for it.

PR #491, opened 2019-05-18
add yarn pnp support

lroling8350:
I am using yarns new [PnP](https://yarnpkg.com/en/docs/pnp) node resolution. Unfortunately husky does not work with PnP because it starts node without pnp resolution. This PR is a simple way on install to detect that pnp is enabled by requiring pnpapi. If it exists add the --require to the pnp file as NODE_OPTIONS to be picked up by node.

### DIFF
--- a/src/installer/__tests__/__snapshots__/getScript.ts.snap
+++ b/src/installer/__tests__/__snapshots__/getScript.ts.snap
@@ -35,6 +35,7 @@ if [ -f \\"$scriptPath\\" ]; then
     debug \\"source ~/.huskyrc\\"
     . ~/.huskyrc
   fi
+  
   node_modules/run-node/run-node \\"$scriptPath\\" $hookName \\"$gitParams\\"
 else
   echo \\"Can't find Husky, skipping $hookName hook\\"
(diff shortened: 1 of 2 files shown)


lroling8350:
I wanted to write some additional unit tests but i can't for the life of me figure out how to mock the require.resolve function.

typicode:
Hi @lroling8350,

Thanks for the PR.

There seems to be a `NODE_OPTIONS`  set by yarn. It can be seen by adding a `preinstall`  hook for example in `package.json`:

```json
"scripts": {
  "preinstall": "printenv"
}
```

```sh
yarn --pnp
```

I suppose it could be used when generating husky script (https://github.com/typicode/husky/blob/master/src/installer/getScript.ts#L71)

Something like:

```js
const nodeOptions = /* ... */ process.env.NODE_OPTIONS

NODE_OPTIONS=${nodeOptions} $[node} ...
```

The env could then be mocked.

lroling8350:
Ahhh, that would be a smart way to allow for testing.

My one concern would be if other NODE_OPTIONS are added by the user which would then be added the created hook. This could lead to either options that shouldn't exist during calls or duplicates which could cause errors.

In the code you've provided, it's also not account for node_options the user may want to pass in as they are overwritten with NODE_OPTIONS. 

I'm think to improving testing I can actually separate out the function checking for yarn pnp in a way that can be mocked.

arcanis:
Oh interesting, I missed this PR! 😮 I've just pushed #511 which would make the whole process go through Yarn. I think it's the best option, as it gives all latitude to the package manager to decide how to setup the environment to run the dependencies.

PR #504, opened 2019-06-13
Support searchParentDirs property ("root" property)

### LINKED ISSUE #502: Question: Why does husky findUp git?
I'm breaking this out from #499

I've got a project whose nature requires having multiple mini-projects ("templates") in a directory of the project, all of which are committed to git.

When husky installs, it uses `find-up` to look up recursively until it finds a parent folder w/ a `.git` file or folder:

https://github.com/typicode/husky/blob/8b9989f0696e0f0c330991ebe46e93181d2f9ad8/src/installer/index.ts#L138

This means that if you install husky in `root/a/b/c`, it'll set `root/.git` to call husky in `root/a/b/c`.

My question is: Why?

Personally, I'd think you'd want to just check if the `package.json` root had a `.git` file or folder (so `/c` in this example); otherwise you're running scripts that are hidden away in nested folders, and also need to make sure the packages are actually installed if they're critical to your project (or it's flow).

It could be I've completely missed a valid use case, hence the question :)

G-Rath:
This closes #502 and should generally make life easier for people wanting to install husky multiple times in the same repo.

Originally, I proposed `root` as it's based off the same property in [`eslint`s config](https://eslint.org/docs/user-guide/configuring#using-configuration-files-1).

However, after discovering that `husky` only supports configs in the same directory as the installing `package.json` I felt `searchParentDirs` was a better fit.

It's `true` by default.

### DIFF
--- a/src/installer/__tests__/index.ts
+++ b/src/installer/__tests__/index.ts
@@ -318,6 +318,37 @@ describe('install', (): void => {
     expect(exists('.git/hooks/pre-commit')).toBeFalsy()
   })
 
+  it('should not install hooks if searchParentDirs is false & cwd is git-less', (): void => {
+    mkdir('.git/hooks')
+    mkdir('A/B/node_modules/husky')
+    writeFile(
+      'A/B/package.json',
+      JSON.stringify({ husky: { searchParentDirs: false } })
+    )
+
+    installFrom('A/B/node_modules/husky', 'A/B/node_modules/run-node/run-node')
+
+    expect(exists('.git/hooks/pre-commit')).toBeFalsy()
+  })
+
+  it('should install hooks if searchParentDirs is false & cwd is a git directory', (): void => {
+    const huskyDir = 'node_modules/husky'
+    const hookFilename = '.git/hooks/pre-commit'
+
+    mkdir('.git/hooks')
+    mkdir(huskyDir)
+    writeFile(
+      'package.json',
+      JSON.stringify({ husky: { searchParentDirs: false } })
+    )
+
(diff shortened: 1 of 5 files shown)


src/installer/__tests__/index.ts
```diff
@@ -318,6 +318,37 @@ describe('install', (): void => {
     expect(exists('.git/hooks/pre-commit')).toBeFalsy()
   })
 
+  it('should not install hooks if searchParentDirs is false & cwd is git-less', (): void => {
```
G-Rath: Typically I'd use a `describe` to group these tests, but didn't know what style you prefer (given lack of `describe` & use of `should`).

G-Rath:
#508 changes how `husky` finds the `.git` directory by using `git rev-parse` instead of `find-up`.

On the downside, this PR will pretty much need to be reimplemented.
On the upside, the new implementation is pretty much a couple of lines:

```
if (conf.root && path.resolve(userPkgDir) !== path.resolve(gitDir)) {
  console.log(
    "Can't find .git in root directory, skipping Git hooks installation."
  )
  return
}
```

This would go below this [line](https://github.com/typicode/husky/pull/508/files#diff-a6471f9b9a786e85a8d013f1adc62112R139); also note the property name change, as `root` becomes a better fit.

I'll rebase & update this PR once #508 is merged.

typicode:
Thanks for the PR, #508 has been merged but I'd suggest to wait a little bit before reworking/updating it.

G-Rath:
Superseded by #521

PR #510, opened 2019-06-25
Adds support for repository-local huskyrc files

arcanis:
This PR adds support for repository-local huskyrc files.

I moved the execution order because of a separate PR I'm working on.

### DIFF
--- a/src/installer/__tests__/__snapshots__/getScript.ts.snap
+++ b/src/installer/__tests__/__snapshots__/getScript.ts.snap
@@ -15,6 +15,8 @@ exports[`hookScript should match snapshot (OS X/Linux) 1`] = `
 
 scriptPath=\\"node_modules/husky/run.js\\"
 hookName=\`basename \\"$0\\"\`
+
+gitRoot=\\"$(git rev-parse --show-toplevel)\\"
 gitParams=\\"$*\\"
 
 debug() {
@@ -22,6 +24,15 @@ debug() {
     echo \\"husky:debug $1\\"
   fi
 }
+
+if [ -f ~/.huskyrc ]; then
+  debug \\"source ~/.huskyrc\\"
+  . ~/.huskyrc
+fi
+ if [ -f \\"\${gitRoot}\\"/.huskyrc ]; then
+  debug \\"source \${gitRoot}/.huskyrc\\"
+  . \\"\${gitRoot}\\"/.huskyrc
+fi
 
 debug \\"$hookName hook started\\"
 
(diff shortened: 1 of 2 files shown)


arcanis:
It seems the current design conflicts with this paragraph from the documentation:

> Starting with `1.0.0`, you can also configure hooks using **`.huskyrc`**, `.huskyrc.json` or `.huskyrc.js` file.
>
> ```js
> // .huskyrc
> {
>   "hooks": {
>     "pre-commit": "npm test"
>   }
> }
> ```

What do you think, @typicode? I guess an alternative would be to use `.huskyrc.sh` for the repository-level rc file, rather than the regular `.huskyrc` that's used for `$HOME`. It might be a bit confusing, but I don't see how to make it work otherwise apart from bumping the major and removing the documented behavior for `.huskyrc`.

arcanis:
@typicode do you think this change could land in some form? I currently need to ensure that each engineer in my org has a variable set in their environment before they can use husky; would be nice to configure it directly from the repo! 😃

voxpelli:
Seems like there has been an unintentional conflict in this module between the `cosmiconfig` based setup added in #209 and the sourcing of `.huskyrc` that was added in #397

In a previous version of the documentation it was however noted that it did differ from the `cosmiconfig` one: https://github.com/typicode/husky/commit/2f0228a260bf6f8d4bcbb310fb67f4af22584d82#diff-54ca93aaacd5ce53520f97f106d6dca4L82

I would suggest not following any of the `.huskyrc.*` as that's pretty much cosmiconfig domains in this case: https://www.npmjs.com/package/cosmiconfig Would be better to clearly separate this.

I would push for something like `.huskyenv`, `.huskyconfig`, maybe `.huskysh` or such instead. Maybe even `.huskyenv.sh`?

I would also propose that, at the same time, changing the filename in `$HOME` to match that new name and deprecating the old name. Reason is that the old name is also used by `cosmiconfig` if husky is used in a subfolder to `$HOME`, thus causing a possible conflict between the two.

As the feature is marked as experimental, such a change could be made without having to increase a major version, but a simple deprecation which prints a warning would be just as easy, so I guess that would be preferable :)

arcanis:
@typicode is it fine if I change the file name to `.huskysh`? Do you want to also support this name for the home file, for consistency?

typicode:
Regarding naming, I don't really know what would be the best.

I agree that `$HOME/.huskyrc` and `project/.huskyrc` are confusing right now.

Cosmiconfig could be restricted to only `project/.husky.config.js` and `.huskyrc` would be free to be used for running shell commands. It would also match other `$HOME/.bashrc`, `$HOME/.zshrc`, ... files which contain shell commands.

Alternatively, `.husky.sh` or `.huskyenv` are nice too and as said wouldn't conflict with cosmiconfig. But one use case for `$HOME/.huskyrc` is to be able to run/load `nvm` for GUIs, I don't know if `env` suffix would be meaningfull there, maybe.

It makes sense in `project/.huskyenv` though.

I'm going to merge this PR to the `next` branch and work on it from there.

arcanis:
@typicode By any chance, do you have an approximate idea of when you'd see this (or a similar feature) reaching master?

PR #521, opened 2019-06-28
Support root property

### LINKED ISSUE #502: Question: Why does husky findUp git?

G-Rath:
This closes #502 and should generally make life easier for people wanting to install husky multiple times in the same repo.

Superseeds #504

It's based off the `root` property in [`eslint`s config](https://eslint.org/docs/user-guide/configuring#using-configuration-files-1).

It's `false` by default, and should work w/ git-submodules, but I don't know about working trees as I've not used either.

### DIFF
--- a/DOCS.md
+++ b/DOCS.md
@@ -71,6 +71,38 @@ Generally speaking, you should AVOID defining `husky` in multiple `package.json`
 }
 ```
 
+Alternatively, you can use the `root` property to tell `husky` to only install if the [git `toplevel`](https://git-scm.com/docs/git-rev-parse#Documentation/git-rev-parse.txt---show-toplevel) directory is the same as the one with the installing `package.json`:
+
+```sh
+.
+└── root 🐶 # git toplevel points here
+    ├── .git
+    ├── package.json # husky will be installed as a depenceny *and* hooks will be installed
+    └── packages
+        ├── A
+        │   └── package.json # husky will be installed as a depenceny *only* (i.e hooks will not installed)
+        ├── B
+        │   └── package.json # husky will be installed as a depenceny *only* (i.e hooks will not installed)
+        └── C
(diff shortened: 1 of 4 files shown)


DOCS.md
```diff
@@ -71,6 +71,38 @@ Generally speaking, you should AVOID defining `husky` in multiple `package.json`
 }
 ```
 
+Alternatively, you can use the `root` property to tell `husky` to only install if the [git `toplevel`](https://git-scm.com/docs/git-rev-parse#Documentation/git-rev-parse.txt---show-toplevel) directory is the same as the one with the installing `package.json`:
+
+```sh
+.
+└── root # git toplevel points here
+    ├── .git
+    ├── package.json 🐶 # husky will be installed as a depenceny *and* hooks will be installed
```
vsimon: spelling: dependency
\+ in a few other places below

G-Rath: So it seems IntelliJ doesn't spell check markdown sh code blocks?

Thanks for catching this!

G-Rath:
@typicode this is ready to go when you are :)

PR #539, opened 2019-07-12
Add HUSKY_SKIP_UNINSTALL option

damccorm:
I think it would be useful to have the option to not uninstall git hooks when husky is uninstalled.

For my use case, I need to check in my node_modules, but I don't want to check in dev-dependencies. So during my post-commit stage, I'm running something like:

`npm prune --production && git add node_modules/* && git commit -m \"Husky commit correct node modules\" && npm install`

In general, this works great - the correct node_modules get committed - however, when I run npm install Husky tries and fails to add its hooks since the hook file its trying to write to is already open. So I lose my hooks in the process. What I'd really like to do is

`export HUSKY_SKIP_UNINSTALL=1 && export HUSKY_SKIP_INSTALL=1 && npm prune --production && git add node_modules/* && git commit -m \"Husky commit correct node modules\" && npm install`

### DIFF
--- a/DOCS.md
+++ b/DOCS.md
@@ -28,6 +28,12 @@ If you don't want `husky` to automatically install Git hooks, simply set `HUSKY_
 HUSKY_SKIP_INSTALL=1 npm install
 ```
 
+Likewise, if you don't want `husky` to automatically uninstall the hooks it has previously installed, set `HUSKY_SKIP_UNINSTALL` environment variable to `1`.
+
+```sh
+HUSKY_SKIP_UNINSTALL=1 npm install
+```
+
 ## Skip all hooks
 
 During a rebase you may want to skip all hooks, you can set `HUSKY_SKIP_HOOKS` environment variable to `1`.
--- a/src/installer/__tests__/index.ts
+++ b/src/installer/__tests__/index.ts
@@ -303,7 +303,29 @@ describe('install', (): void => {
     expect(hook).toMatch(huskyIdentifier)
   })
 
-  it('should not install hooks if HUSKY_SKIP_INSTALL=1', (): void => {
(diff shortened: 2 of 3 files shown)


patrickmichalina:
If this branch is fixed up will it be considered for merge?

patrickmichalina:
Really no reason to see this sort of message on CI servers:

```
-----> Pruning devDependencies
       
       > husky@4.2.5 preuninstall node_modules/husky
       > node husky uninstall
       
       husky > Uninstalling git hooks
       fatal: not a git repository (or any parent up to mount point /)
       Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
       husky > Failed to uninstall
       removed 1118 packages in 20.259s

```

damccorm:
> If this branch is fixed up will it be considered for merge?

Happy to do the work here, but I'd like to have an indication that this change will be merged before sinking more time into it

mmcxii:
Curious what the status of this is. I am currently facing in issue with optimizing a Docker image and uninstalling unused dependencies, however Husky is causing that step to fail because it can't be uninstalled in the container.

Here's my terminal output from attempting to uninstall Husky

```
[1/2] Removing module husky...
$ node husky uninstall
husky > Uninstalling git hooks
git rev-parse command failed. Got null
husky > Failed to uninstall
$ node husky uninstall
husky > Uninstalling git hooks
git rev-parse command failed. Got null
husky > Failed to uninstall
[2/2] Regenerating lockfile and installing missing dependencies...
```

damccorm:
Still happy to fix this up if I get the go-ahead from @typicode or another maintainer, but I'm not actively using the tool anymore so I'm not going to make too much of a push for it

mmcxii:
Is there another tool that's taken its place in your stack? I'd hate to lose out on being able to automatically run commitlint and lint-staged.

PR #535, opened 2019-07-08
Optionally overwrite existing hooks

tomasbjerre:
# Problem

I'm working within an organization where we get Git Bash pre packaged with hooks. We cannot change that and every developer will get these hooks, from the installation of Git Bash, installed when cloning repositories. The pre packaged hooks will make Husky ignore any hooks configured in `package.json`.

Developers must do something like:
```shell
git clone whatever.git
rm ./git/hooks/*
npm install
```

# Solution

Adding a configuration option (`overwriteExisting`) to `package.json` like:
```json
...
"skipCI": true,
"overwriteExisting": true
"pre-commit": ...
``` 

With `overwriteExisting` as `true` in the configuration it will overwrite any existing hook on `npm install`.

Fixes #336 and #558

PR #559, opened 2019-08-12
Only install/uninstall hooks specified in config

tomasbjerre:
Installing all the hooks by default can make things like rebasing super slow. This patch makes the installer only manage the hooks specified in the config.

This is a rebase of https://github.com/typicode/husky/pull/204 by @xdissent because that PR has merge conflicts.

### DIFF
--- a/src/installer/__tests__/index.ts
+++ b/src/installer/__tests__/index.ts
@@ -15,7 +15,9 @@ const defaultGitDir = '.git'
 const defaultGitHooksDir = '.git/hooks'
 const defaultHookFilename = '.git/hooks/pre-commit'
 const defaultHuskyDir = 'node_modules/husky'
-const pkg = JSON.stringify({})
+const pkg = JSON.stringify({
+  husky: { hooks: { 'pre-commit': 'echo' }, skipCI: true }
+})
 
 // Helpers
 function install({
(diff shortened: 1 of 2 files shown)


typicode:
Hi @tomasbjerre,

Thanks for the PR :)

For rebasing, I would suggest using `HUSKY_SKIP=1` which has been added for this particular case (https://github.com/typicode/husky/blob/master/DOCS.md#skip-all-hooks).

I agree that faster is better but for the moment I'd prefer to keep the current way as it would come with trade-offs.

The concern I have with installing only hooks in config is that it may make installation a little less user-friendly for beginners.

For example, this would fail:

```sh
# First install
npm install husky --save-dev

# Add "pre-commit": "npm test" to .huskyrc
# ...

git commit -m foo
# No test would run because pre-commit wasn't yet defined in .huskyrc
```

Or if a `commit-msg` hook is added later, `npm install` would need to be run again or else it won't work.

Also, by default hooks shouldn't add that much to git commands (I tested on a budget laptop/Linux and it's less than `0.5s` for a git commit). In my experience, what makes things slow is how many things are done in hooks (linting, testing, etc...).

You can check how much time it adds with the following commands. Let me know if it's more than `0.5s`.

```sh
git init
npm init -y && npm install husky --save-dev
touch foo && git add foo
time HUSKY_DEBUG=1 git commit -m foo
```

tomasbjerre:
Ok @typicode . Would be nice if you can also have a look at my other PR: https://github.com/typicode/husky/pull/535

max-mykhailenko:
How I can help with testing? I think that a lot of delays happens because we have many hooks

casz:
perhaps as part of the install of husky it could warn if no hooks were found in any of the config file and basically fail the install?

yinzara:
Any love?

typicode:
Fixed in v5. Thanks again for the PR :+1:

PR #568, opened 2019-08-26
Next.js doesn't use husky now

sheerun:


typicode:
Thanks for the update 👍

Issue #565, opened 2019-08-21
husky package post install hook fails on windows


capaj:
```
> husky@3.0.3 postinstall C:\Users\jspac001\repos\ProductivityHubUI\node_modules\husky
> opencollective-postinstall || true

'opencollective-postinstall' is not recognized as an internal or external command,
operable program or batch file.
'true' is not recognized as an internal or external command,
operable program or batch file.
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! husky@3.0.3 postinstall: `opencollective-postinstall || true`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the husky@3.0.3 postinstall script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

```
is what happens on a windows machine.

typicode:
Thanks for the report. I'm not sure why it can't find `open-collective`. It's defined as a dependency and I've just tried on Windows.

Could you try with a freshly created repo?

```sh
git init
npm init -y
npm install husky --save-dev
```

capaj:
I confirm that clean installs run fine.

typicode:
@emily-curry thanks a lot for checking the other projects, do you want to make a PR with `|| exit 0`?

naingaungphyo:
Now latest(master) core-js code changed from 
> || operator

 to 
> "postinstall": "node -e \"try{require('./postinstall')}catch(e){}\"".

Husky postinstall script should be changed into the same way, because it is causing error in powershell. 
The error is the same as this [core-js issue](https://github.com/zloirock/core-js/issues/582)

PR #573, opened 2019-08-28
[Fix] Prevent postinstall from failing on windows

emily-curry:
Fixes: https://github.com/typicode/husky/issues/565

From that issue:
> On windows, there's no `true`. Looking at other popular packages, I believe `|| exit 0` is a more reliable way to handle this ([1](https://github.com/remy/nodemon/blob/master/package.json), [2](https://github.com/styled-components/styled-components/blob/master/packages/styled-components/package.json))

This change was tested by removing the `opencollective-postinstall` script from the .bin folder, and then running `npm run postinstall` in cmd and powershell. In both shells I validated that without this change, the process exits with a non-zero exit code (cannot find 'opencollective-postinstall', cannot find 'true'), and with this change that it exits with an exit code of zero.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "postpublish": "pinst --disable",
     "lint": "eslint . --ext .js,.ts --ignore-path .gitignore",
     "fix": "npm run lint -- --fix",
-    "postinstall": "opencollective-postinstall || true"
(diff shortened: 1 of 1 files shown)


typicode:
Thanks!

PR #576, opened 2019-09-02
Update README.md

typicode:


package.json
```diff
@@ -22,7 +22,7 @@
     "postpublish": "pinst --disable",
     "lint": "eslint . --ext .js,.ts --ignore-path .gitignore",
     "fix": "npm run lint -- --fix",
-    "postinstall": "opencollective-postinstall || exit 0"
+    "_postinstall": "opencollective-postinstall || exit 0"
```
armano2: @typicode q: is this a typo?, if not this hook will not be triggered

typicode: It’s prefixed with a _ to prevent postinstall to run in development. But it’s renamed to postinstall by pinst during prepublish.

But I can understand it’s confusing :)

PR #577, opened 2019-09-04
Expose husky-installer as an executable

fastfrwrd:
We found ourselves wanting to implement a custom setup where we need to re-run husky's installer if we find the wrong git-hook in place post-install. To do this, we are having to run `./husky.js` directly. We think it would make sense to expose this to users in our situation.


mbrevda:
Would `npm rebuild husky` work form you?

fastfrwrd:
yes that actually works great! I'll close the PR; I hope yall will consider documenting that that's a good way to fix any issues that arise if your hook files were modified unexpectedly.

PR #579, opened 2019-09-09
Remove is-ci in favor of ci-info

xenyal:
Instead of having is-ci call ci-info's ci.isCI, we might as well call ci.isCI within the installation directly. This allows us to shorten the dependency branch by 1 step which is a trivial amount of gain, but it nonetheless reduces the length of the package-lock

### DIFF
--- a/src/installer/bin.ts
+++ b/src/installer/bin.ts
@@ -1,5 +1,5 @@
 import chalk from 'chalk'
-import isCI from 'is-ci'
+import { isCI } from 'ci-info'
 import path from 'path'
 import debug from '../debug'
 import { install, uninstall } from './'
--- a/package-lock.json
+++ b/package-lock.json
@@ -502,6 +502,11 @@
         "@babel/types": "^7.3.0"
       }
     },
+    "@types/ci-info": {
+      "version": "2.0.0",
+      "resolved": "https://registry.npmjs.org/@types/ci-info/-/ci-info-2.0.0.tgz",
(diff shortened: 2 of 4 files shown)


typicode:
Thanks!

src/installer/bin.ts
```diff
@@ -13,6 +13,9 @@ debug(`INIT_CWD: ${process.env.INIT_CWD}`)
 // huskyDir is ONLY used in dev, don't use this arguments
 const [, , action, huskyDir = path.join(__dirname, '../..')] = process.argv
 
+// Determine whether environment is CI
+const { isCI } = ci.isCI
```
typicode: Except if I'm missing something, shouldn't it be `const { isCI } = ci` based on https://github.com/watson/ci-info#usage?

xenyal: Based on the library's usage documentation, `ci` is the encapsulating namespace of the library, and `isCI` is the method that actually shows whether or not the call was invoked in a CI environment or not. Not sure if that answers your question, I'll keep an 👀out for further comments! Also, sorry for the late response!

fearphage:
```suggestion
const { isCI } = ci
```

Looks correct to me based on the linked documentation.

xenyal: @typicode @fearphage Updated - thank you both for the replies! 😄

src/installer/bin.ts
```diff
@@ -1,5 +1,5 @@
 import chalk from 'chalk'
-import isCI from 'is-ci'
+import ci from 'ci-info'
```
fearphage: More of a preference really, but just pull this out during import?

```suggestion
import { isCI } from 'ci-info'
```

xenyal: I like it, updated!

typicode:
Thanks @fearphage, @xenyal

PR #625, opened 2020-01-06
Merge 4.0

typicode:


christianbundy:
Hi! Sorry to bother you, but what are the breaking changes between 4.0.0 and 3.x.x? I can't find them anywhere and I'm unsure about whether I should upgrade.

typicode:
Hi @christianbundy,

No worries. They can be found here https://github.com/typicode/husky/releases

The only breaking change is that Husky requires `Node 10+` instead of `Node 8+`. The other changes don't require any update from the user 👍

zuzusik:
@typicode the other breaking change was moving from `run-node` to  NPX which made Husky majorly broken in many GUI git clients when upgrading from 3.x to 4.x

PR #633, opened 2020-01-09
fix: use process.exit if install fails

merceyz:
When using Yarn v2 there is no console output from installs unless they fail, since husky doesn't change the exit code it seems like the install succeeded when it actually failed

### DIFF
--- a/src/installer/bin.ts
+++ b/src/installer/bin.ts
@@ -103,6 +103,7 @@ function run(): void {
     console.log(chalk.red(err.message.trim()))
     debug(err.stack)
     console.log(chalk.red(`husky > Failed to ${action}`))
+    process.exit(1)
   }
 }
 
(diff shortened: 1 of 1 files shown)


src/installer/bin.ts
```diff
@@ -103,6 +103,7 @@ function run(): void {
     console.log(chalk.red(err.message.trim()))
     debug(err.stack)
     console.log(chalk.red(`husky > Failed to ${action}`))
+    process.exit(1)
```
arcanis: Better use `process.exitCode`, since `process.exit` might prevent the logs from being flushed:
```suggestion
    process.exitCode = 1;
```

merceyz: Sure 👍 
Remove the semicolon and i'll commit your suggestion.

This repo has semi set to false https://github.com/typicode/husky/blob/0ffe7853853dd7667ddc24a78469b92b5312aeb2/.prettierrc#L2

merceyz: Sure 👍 
Remove the semicolon and i'll commit your suggestion.

This repo has semi set to false https://github.com/typicode/husky/blob/0ffe7853853dd7667ddc24a78469b92b5312aeb2/.prettierrc#L2

arcanis: Woops, updated 👍

typicode:
Thanks!

PR #653, opened 2020-01-20
fix snapshot

casz:
seems like sh on appveyor is using WSL's sh meaning you might as well be testing on Linux 😆


typicode:
> seems like sh on appveyor is using WSL's sh meaning you might as well be testing on Linux laughing
Good to know :+1:

casz:
Due note my above statement is only relevant for appveyor when using Visual Studio 2019 image

PR #692, opened 2020-03-21
replace postinstall with npm funding

ChristianMurphy:
The use of post install scripts is somewhat controversial.
This blog gives some good context https://feross.org/funding-experiment-recap
Using just `npm fund` preserves the behavior of alerting users that there is an opportunity to contribute financially, without some of the drawbacks of running arbitrary scripts in `postinstall` comes with.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6191,11 +6191,6 @@
         "mimic-fn": "^2.1.0"
       }
     },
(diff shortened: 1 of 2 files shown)


ajgrover:
out of curiosity, why was this closed?

PR #701, opened 2020-04-01
Show runtime node version when exit

zhangciwu:
It's useful when some git GUI apps can't run `node -v` directly


typicode:
Thanks :+1:

PR #726, opened 2020-05-14
Update README.md - Add "How it works"

iwt-nduesing:


mgan59:
This was my immediate question -- "How does this thing work"?  But more so, I understand there are underlying `hooks`.  But what I want to understand is how/when the hooks get hoisted into place.

With that said, this proposed update is step in the right direction.  But needs just a little more meat/potates.  just my thoughts.

typicode:
Thanks for the PR/feedback, good explanation :+1: 
However, I think I'll leave docs for v4 how they are.

With v5, I tried to make husky simpler so I hope it to be easier to understand :)

iwt-nduesing:
Even if something is simple, it's not necessarily easy to understand. I like how volta explained how it works: https://docs.volta.sh/guide/#how-does-it-work I would appreciate a description not just how to use this tool but also in how it works internally. That helps to understand whether there are possible performance issues, security risks and things like that.

Issue #677, opened 2020-02-21
Husky setting wrong folder in monorepo (lerna/yarn workspace)


tad3j:
Today I've noticed that Husky hooks suddenly stopped working in a monorepo. After investigating I've found out that file `husky.local.sh` inside `.git\hooks` had an entry on last line `cd "packages/shared/"`. After removing this line Husky hooks started to work again.

My folder structure is like:
root:
 -package.json (with Husky hooks setup)
 -.git
 --packages
 ---shared
 -----package.json
 ---web
 ------package.json

Could this happen when I ran `yarn add/install` or `lerna clean && lerna bootstrap` from a sub-directory? Is Husky aware of lerna/yarn workspaces? One solution to always use the right directory would be to check package.json for `husky` key.

OS: Windows 10
Terminal: webstorm, CMD, cygwin
Since issue was pinpointed I'm not including command's outputs.

bertdeblock:
I'm also using the same setup (Yarn Workspaces + Lerna) and I've noticed the exact same issue.

yuanguandong:
i have the same issue

melanieseltzer:
My setup is almost identical (using Yarn workspaces only, no Lerna) and I'm getting the same. This is what my structure looks like:

```
package.json (Husky setup at root)
- packages
  - client
    package.json
  - server
    package.json
```
I just did some testing and it's only happening with the `yarn remove ...` command, and not with `yarn add ...`.

Changes the `husky.local.sh` from this (which works fine):

```
packageManager=yarn
cd "."
```

to this (the hook stops running on commit):

```
packageManager=yarn
cd "packages/client/"
```

bertdeblock:
Just tested this again as well and I can confirm it only seems to happen when running `yarn remove ...`.

eduard-malakhov:
There is a special node on using husky with monorepos: https://github.com/typicode/husky#monorepos In short for monorepos, it is recommended to only install husky into the root package. Otherwise, all sorts of conflicts are expected when husky installation script is run for several packages in the repo.

I use husky with lerna and yarn workspaces by installing husky to the root package via `yarn add --dev -W husky`. The `-W`/`--ignore-workspace-root-check` flag allows installation into the root package (https://classic.yarnpkg.com/en/docs/cli/add#toc-yarn-add-ignore-workspace-root-check-w). Personally, I haven't seen any downsides of this approach so far.

tad3j:
[...]

PR #728, opened 2020-05-26
Add flag to fix relativeUserPkgDir

bnbarak:
# What
Per https://github.com/typicode/husky/issues/677, and @eduard-malakhov great explanation https://github.com/typicode/husky/issues/677#issuecomment-602520663

Fix a bug when running `yarn|npm install` from a sub `node_modules` in a monorepo/yarn workspaces project.

**Current:** the relative path to the project's root is set no `packages/someOther` when running from a package like `packages/someOther`.
**Expected**: To be able to force the relative path to the be the root of the project even when the `yarn|npm install` is running not from the root.

### DIFF
--- a/README.md
+++ b/README.md
@@ -328,7 +328,7 @@ npm install husky --save-dev
 
 Verify that your version of Git is `>=2.13.0`.
 
-#### Lerna Workspaces and Monorepos
+#### Lerna, Monorepos, and Yarn Worksapces
 If your project has nested `node_modules` or you run `npm|yarn install` not from the project's root,
 simply set `HUSKY_RUN_FROM_ROOT` environment variable to force husky relative path to be `.`.
 
--- a/README.md
+++ b/README.md
@@ -328,6 +328,10 @@ npm install husky --save-dev
 
 Verify that your version of Git is `>=2.13.0`.
 
+#### Lerna Workspaces and Monorepos
+If your project has nested `node_modules` or you run `npm|yarn install` not from the project's root,
+simply set `HUSKY_RUN_FROM_ROOT` environment variable to force husky relative path to be `.`.
+
 ## See also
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -328,6 +328,10 @@ npm install husky --save-dev
 
 Verify that your version of Git is `>=2.13.0`.
 
+#### Lerna Workspaces and Monorepos
```
bryantee, line 331: Seems like this fix would also address this happening in a Yarn Workspaces monorepo. Maybe we should call that out?

bnbarak: Changed to `Lerna, Monorepos, and Yarn Worksapces` to cover everything. What do you think?

bnbarak: Thanks :) Already updated the PR, what is your merge process?

bryantee: Maybe @typicode can chime in for merge process.

I'd love to test this out @bnbarak, what's the easiest way?

bnbarak: #### First lets recreate the issue on master:
0. close this example of a basic Lerna monorepo https://github.com/reggi/lerna-tutorial
1.  Setup Husky
2. Run `yarn` inside `packages/usage`
3. Open `.git/hooks/huskey.local.sh`, expect to see `cd "packages/usage"` // <-- **this is the problem**

#### Test the fix
4. Checkout this PR in the example project
5. Set the env variable `HUSKY_RUN_FROM_ROOT="true"` or `HUSKY_RUN_FROM_ROOT="1"`
6. Remove `.git/hooks`
7. Run `yarn` inside `packages/usage`
8. Open `.git/hooks/huskey.local.sh`, expect to see `cd "."` // <-- **this is the fix**

bnbarak: Piniging @lessen-inc and @typicode for their attention :)

Faolain:
hey just bumping this, any reason why it hasn't been merged? @typicode @lessen-inc

Faolain:
@typicode is there a reason this was closed and not merged? cc: @bnbarak @bryantee

Issue #735, opened 2020-06-09
HUSKY_GIT_PARAMS not complete at Windows/cygwin


amjmhs:
HUSKY_GIT_PARAMS are not correctly provided to a node-process for prepare-commit-message.

OS: Windows 10
Shell: cygwin
Node: 12.16.1
Husky: 4.2.5

`"prepare-commit-msg": "node bin/prepare-commit-msg/git-commit-template.js"`
```javascript
bin/prepare-commit-msg/git-commit-template.js
---------------------------------------------

#!/usr/bin/env node

console.dir(process.env.HUSKY_GIT_PARAMS);
```

**Output:**
```shell
$ git commit
husky > prepare-commit-msg (node v12.16.1)
'.git/COMMIT_EDITMSG'

$ git commit               (with existing commit.template-setting)
husky > prepare-commit-msg (node v12.16.1)
'.git/COMMIT_EDITMSG'

$ git commit -m "foobar"
husky > prepare-commit-msg (node v12.16.1)
'.git/COMMIT_EDITMSG'

$ git commit --amend
husky > prepare-commit-msg (node v12.16.1)
'.git/COMMIT_EDITMSG'
```

**Expected Output:**
```shell
$ git commit
husky > prepare-commit-msg (node v12.16.1)
'.git/COMMIT_EDITMSG'

$ git commit               (with existing commit.template-setting)
husky > prepare-commit-msg (node v12.16.1)
'.git/COMMIT_EDITMSG template'

$ git commit -m "foobar"
husky > prepare-commit-msg (node v12.16.1)
'.git/COMMIT_EDITMSG message'

$ git commit --amend
husky > prepare-commit-msg (node v12.16.1)
'.git/COMMIT_EDITMSG commit HEAD'
```

Am I doing something wrong here?

I think the root-cause is at the `.git/hooks/husky.sh` - Line 17
`"$@" husky-run $hookName "$gitParams"`
should be
`"$@" husky-run $hookName \"$gitParams\"`

kirkoman:
I'm having an issue I think is related.  In a windows + git bash + npm + husky + lfs environment I was constantly getting "this should be run through git's post-checkout hook" even when I *was* running lfs through the hook:

```
C:\my-repo>git checkout develop
Switched to branch 'develop'
Your branch is behind 'origin/develop' by 1 commit, and can be fast-forwarded.
  (use "git pull" to update your local branch)
husky:debug husky v4.2.5 - post-checkout
[...]

PR #737, opened 2020-06-22
fix prepare-commit-msg on windows

kirkoman:
Should fix issue #735

### DIFF
--- a/sh/husky.sh
+++ b/sh/husky.sh
@@ -10,7 +10,7 @@ command_exists () {
 
 run_command () {
   if command_exists "$1"; then
-    "$@" husky-run $hookName "$gitParams"
+    "$@" husky-run $hookName $gitParams
 
     exitCode="$?"
     debug "$* husky-run exited with $exitCode exit code"
--- a/src/installer/__tests__/__snapshots__/scripts.ts.snap
+++ b/src/installer/__tests__/__snapshots__/scripts.ts.snap
@@ -39,7 +39,7 @@ command_exists () {
 
 run_command () {
   if command_exists \\"$1\\"; then
-    \\"$@\\" husky-run $hookName \\"$gitParams\\"
+    \\"$@\\" husky-run $hookName $gitParams
 
     exitCode=\\"$?\\"
     debug \\"$* husky-run exited with $exitCode exit code\\"
--- a/src/runner/__tests__/index.ts
+++ b/src/runner/__tests__/index.ts
@@ -138,8 +138,7 @@ describe('run', (): void => {
       })
     )
 
-    // 'commit-msg' takes one parameter from git
-    const status = await index(['', '', 'commit-msg', 'git fake param'], {
+    const status = await index(['', '', 'commit-msg', 'git', 'fake', 'param'], {
       cwd: dir,
     })
     expectSpawnSyncToHaveBeenCalledWith(dir, 'echo success', {
--- a/src/runner/index.ts
+++ b/src/runner/index.ts
@@ -74,7 +74,7 @@ function runCommand(
  * @param {promise} options.getStdinFn - used for mocking only
  */
 export default async function run(
-  [, , hookName = '', HUSKY_GIT_PARAMS]: string[],
+  [, , hookName = '', ...HUSKY_GIT_PARAMS]: string[],
   { cwd = process.cwd() }: { cwd?: string } = {}
 ): Promise<number> {
   const oldCommand = getOldCommand(cwd, hookName)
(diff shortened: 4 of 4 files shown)


kirkoman:
Note that the fix (8f05f8a) still works even if you keep the quotes.  I removed the quotes in husky.sh (as a separate commit - afac6c4) because it seemed better to have all environments uniformly using the rest param syntax.

If the husky.sh change seems to disruptive it can be dropped.  For environments where the quoting is working as expected, those arguments will just come through as a single string and the `join()` will pass it through unmolested.

whschultz:
This approach (hand-modified local js to match) is working for me so long as my hooks in `package.json` look like this:

```json
  "husky": {
    "hooks": {
      "pre-commit": ".githooks/pre-commit.sh ${HUSKY_GIT_PARAMS[@]}",
      "post-commit": ".githooks/post-commit.sh ${HUSKY_GIT_PARAMS[@]}",
      "post-checkout": ".githooks/post-checkout.sh ${HUSKY_GIT_PARAMS[@]}",
      "post-merge": ".githooks/post-merge.sh ${HUSKY_GIT_PARAMS[@]}"
    }
  }
```

kirkoman:
Interesting, I am not finding the array expansion syntax to be necessary.

Did you include the `.join(' ')` change on line 87 when making your local js edits?  This should be serializing the array to a string.

kirkoman:
@whschultz I've added some tests that contradict your results.  The shell syntax for array-parameter expansion should not be needed.  Let me know if I missed something though.

The tests also illustrate that the old behavior was wrong for multi-parameter hooks (viz. `prepare-commit-msg`) as they do fail without the fix.

Hoping a maintainer will weigh in some week soon...

kirkoman:
@typicode got any feels on a husky release any time soon?

steelbrain:
Just got bit by this. Applied the patch and it worked well. Can a maintainer please have a look at this?

rikoe:
I can confirm that this test does indeed fix issue #735 with `git-commit-template`. I have been struggling for a while with this issue, thinking there was something wrong with `git-commit-template`, but the issue is that all the git params aren't being passed through correctly by husky.

Is there any plan to approve and merge this any time soon?

@kirkoman is there a workaround for getting `git-commit-template` to work without this change? I can't really use it as is at the moment...

kirkoman:
@rikoe I wasn't aware of that tool until you referenced it.  Taking a closer look yeah I would not expect it to function properly on Windows based on #735 and the need for this PR.

What I do is to apply a this patch locally when I want to work around this issue.  Ultimately it seems related to a platform difference in how shell is parsing arguments so you might find a workaround by playing around with the quotation-escaping on line 17 of husky.sh (but I lost patience looking for a solution with that approach).

```sh
  "$@" husky-run $hookName "$gitParams"
```

kirkoman:
bump

Updated title to reflect that this is a bug fix.

Just realized from the tag history that v5 is in progress, and it seems to take a very different approach to managing the hooks.

@typicode this is still an ongoing nuisance for myself and presumably the others in the thread.  Hoping to get this PR noticed and merged (or at least closed with some feedback).

typicode:
Sorry for the late reply. Out of the curiosity, I added your test to master and it seems like the current version is correctly working (https://travis-ci.org/github/typicode/husky/jobs/749056907 search for `sixth` tests).

So I guess, it doesn't need to be merged?

PS: thanks for improving the tests

kirkoman:
Cool, thanks for the response.  It's a Windows-only bug so the test only fails on Windows.  I just ran them locally on the tip of master again to confirm, this is what it looks like for me:

```sh
---
hook should run and pass all HUSKY_GIT_PARAMS
---

husky:debug husky v4.3.5 - pre-commit
husky:debug Current working directory is /tmp/husky-project
husky:debug pre-commit config not found, skipping hook
husky:debug husky v4.3.5 - prepare-commit-msg
husky:debug Current working directory is /tmp/husky-project
husky > prepare-commit-msg (node v12.13.1)
prepare-commit-msg hook from Husky
husky:debug npx --no-install husky-run exited with 0 exit code
husky:debug husky v4.3.5 - commit-msg
husky:debug Current working directory is /tmp/husky-project
husky:debug npx --no-install husky-run exited with 0 exit code
husky:debug husky v4.3.5 - post-commit
husky:debug Current working directory is /tmp/husky-project
husky:debug post-commit config not found, skipping hook
[master 1d3c28e] sixth msg
 3 files changed, 0 insertions(+), 0 deletions(-)
 create mode 100644 fifth
 create mode 100644 fourth
 create mode 100644 sixth
Fail: HUSKY_GIT_PARAMS weren't set correctly
.git/COMMIT_EDITMSG != .git/COMMIT_EDITMSG message

C:\Users\kirkoman\Source\husky>
```

I think it is a difference in how the shell escaping or argument parsing works on Windows.  The `gitParams` are wrapped in quotes in the husky.sh which seems to encode all the params as a single string - on all other platforms besides Windows :).

kirkoman:
rebased to fix merge conflict and correct my commiter info

kirkoman:
@typicode note that your appveyor windows build _did_ fail that test: https://ci.appveyor.com/project/typicode/husky/builds/36785480

typicode:
I see, thanks for the details. I'll have another look.

typicode:
Just gave it a try on Windows, really weird... thanks for catching it and finding a fix! :+1:

PR #739, opened 2020-06-26
Fix "huskyrc.yml" -> ".huskyrc.yml"

Thomazella:
https://github.com/typicode/husky/blob/4e1d440be7e3058036d24e7c5558c09448772af2/sh/husky.sh#L38


typicode:
Thanks!

PR #747, opened 2020-07-28
Add per-project global hook

castarco:
## Description

This PR adds a mechanism to source project-specific scripts before executing any of the git hook scripts managed by Husky.

## Rationale
As stated in this comment ( https://github.com/typicode/husky/issues/385#issuecomment-665041037 ), some times is necessary to apply the `exec < /dev/tty` trick before reaching the point where we execute any NodeJS-related tool (in my specific case, because those tools are containerized).

Although it's possible to achieve that by using the `~/.huskyrc` file, using this solution forces all developers in a project to customize their development environments, when, precisely, tools such as `nvm`, or in my specific case `avatar-cli`, are designed to ease the creation of more homogeneous development environments.

Regarding why the file `.huskyhook` file is loaded before `~/.huskyrc`, I assumed that, whatever customization the developer has in its environment, it's probably there because she wants to override something, so I tried to preserve that.

Regarding the name of the file, I'm not attached to it, we could pick something else. I just wanted to use a name different than `.huskyrc` because that name has a different meaning when it's inside a project (Actually I find quite unfortunate that the same name is used for so different concepts: script and configuration).

Related to issue #385 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,6 +249,12 @@ export NVM_DIR="$HOME/.nvm"
 [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
 ```
 
+### Per-Project commands (/project/dir/.huskyhook)
+
+Husky will source `/project/dir/.huskyhook` file if it exists before running hook scripts (also before the local command
+defined in `~/.huskyrc`). You can use it the same way you use `~/.huskyrc`, but being able to track the script with
+`git` and share it with your project colleagues.
+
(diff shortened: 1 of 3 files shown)


rawc0der:
Nice job @castarco. This is going to be really handy.
Any progress about adopting this @typicode  ??

castarco:
@rawc0der it seems that @typicode opted for a simpler approach, in v5 my patch is not as necessary as before.

PR #774, opened 2020-10-16
fix: return non 0 exit code if an error occurred

andreialecu:
Was originally fixed in https://github.com/typicode/husky/pull/633 but it doesn't seem to have survived the update to v5.

Non 0 exit code is required, otherwise Yarn 2 users may not see what went wrong.

### DIFF
--- a/src/bin.ts
+++ b/src/bin.ts
@@ -48,6 +48,6 @@ try {
     help()
   }
 } catch(err) {
-  console.error(err);
-  process.exitCode = 1;
+  console.error(err)
+  process.exitCode = 1
 }
--- a/src/bin.ts
+++ b/src/bin.ts
@@ -30,19 +30,24 @@ Examples
 `)
 }
 
-if (arg === 'add') {
-  add({
-    cwd: process.cwd(),
-    hookName: params[0],
-    cmd: params[1],
-  })
-} else if (arg === 'install') {
-  install({
-    cwd: process.cwd(),
-    dir: params[0],
-  })
-} else if (['--version', '-v'].includes(arg)) {
-  version()
-} else {
-  help()
+try {
+  if (arg === 'add') {
+    add({
+      cwd: process.cwd(),
+      hookName: params[0],
+      cmd: params[1],
+    })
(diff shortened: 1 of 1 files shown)


typicode:
Hi,

Thanks for the PR. Not sure it's still needed.

```js
// index.js
throw new Error('foo')
```

```json
{
 "scripts": {
    "postinstall": "node index.js"
  }
}
```

```sh
❯ yarn -v && yarn run postinstall
2.3.3
/tmp/test/index.js:1
throw new Error('foo')
^

Error: foo
    at Object.<anonymous> (/tmp/test/index.js:1:7)
    at Module._compile (internal/modules/cjs/loader.js:1063:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1092:10)
    at Module.load (internal/modules/cjs/loader.js:928:32)
    at Function.external_module_.Module._load (/tmp/test/.pnp.js:4738:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:72:12)
    at internal/main/run_main_module.js:17:47
```

That said, I think that Yarn v2 doesn't run automatically "postinstall", you can check with `"postinstall": "touch foo"`.

andreialecu:
Yarn 2 does run `postinstall` but only when changes are detected, and it suppresses the output by default. 

Try `CI=true yarn rebuild` and you can test it with 

```
{
 "scripts": {
    "postinstall": "echo hello"
  }
}
```

Regarding why it is needed. It runs it on `yarn install`, but it will not report any error unless you run it manually like in your example.

> Note for Yarn 2 users postinstall won't run automatically, please run yarn run postinstall manually to enable hooks.

This mention in the documentation is not accurate, considering the above.

typicode:
> Yarn 2 does run postinstall but only when changes are detected, and it suppresses the output by default.

Thanks for the explanation, I've updated docs.

typicode:
Closing as I think it's redundant, `node index.js` in the example above already exits with 1. But thanks for spotting the issue in the docs.

PR #824, opened 2020-12-07
fix(shell): double quote params

dargmuesli:
I've set `test/default.sh`'s shebang to `bash` because `echo -e` is used in this file.


typicode:
Thanks for the PR.
What would be the use case for symlinked hooks?

dargmuesli:
Any application that parses the git hooks directory itself without respecting a custom directory set through git config, like GitKraken for example. I suggested respecting the config setting to them earlier today, but until then the many users of GitKraken won't see husky v5 hooks executed as long as they don't add symlinks to the default git hook directory.

dargmuesli:
Why is the CI failing btw? :0

dargmuesli:
I resolved the merge conflicts. @typicode what's your status on this PR? :)

typicode:
Hmm, I'd prefer to keep the script as it is and not rely on `readlink`. I'd rather not have fixes for external tools in husky. 

However, if users want to edit their `pre-commit` hook to handle specifically some tools, they can edit this line manually.

As for the rest LGTM, good catch on the `"$0"` :+1:

.husky/commit-msg
```diff
@@ -1,4 +1,6 @@
 #!/bin/sh
```
typicode, line 1: Could your revert this file and `.husky/pre-commit`

scripts/pre-commit
```diff
@@ -4,7 +4,7 @@ echo "husky: installing Git hooks"
 
 mkdir -p "$husky_dir/_"
 echo "_" > "$husky_dir/.gitignore"
-cp "$(dirname $0)/husky.sh" "$husky_dir/_"
+cp "$(dirname "$0")/husky.sh" "$husky_dir/_"
```
typicode, line 7: :+1:

src/commands/add.ts
```diff
@@ -15,7 +15,12 @@ function createHookFile(file: string, cmd: string) {
     throw new Error(`${file} already exists`)
   }
 
-  const data = ['#!/bin/sh', '. "$(dirname $0)/_/husky.sh"', '', cmd].join('\n')
+  const data = [
```
typicode, line 18:
```suggestion
    const data = ['#!/bin/sh', '. "$(dirname "$0")/_/husky.sh"', '', cmd].join('\n')
```

test/sub-dir.sh
```diff
@@ -1,4 +1,4 @@
-. $(dirname $0)/_functions.sh
+. $(dirname "$0")/_functions.sh
```
typicode, line 1: :+1: could you change `test/config-dir.sh` too?

test/default.sh
```diff
@@ -1,4 +1,7 @@
-. $(dirname $0)/_functions.sh
+#!/bin/bash
```
typicode, line 1: I think the shebang can be removed since these scripts are not executable.

test/default.sh
```diff
@@ -1,4 +1,7 @@
...
+
+# shellcheck source=./_functions.sh
```
typicode, line 3: Is this necessary?

.husky/commit-msg
dargmuesli: I've removed the `readlink` now, or do you want me to revert everything? If yes, why?

test/sub-dir.sh
dargmuesli: Done.

src/commands/add.ts
dargmuesli: Done, except for the additional indentation (?) :)

test/default.sh
dargmuesli: Replaced that with the shellcheck comment.

dargmuesli:
> I'd rather not have fixes for external tools in husky.

I don't really see it as a fix for external tools but as broadened functionality for husky even if the "fix" happened to be the reason for this PR.

typicode (approved):
Thanks! :)

PR #832, opened 2020-12-16
docs: `pre-commit` -> `commit-msg` for commitlint

oebrab:
I've tried add `commitlint` command to `.husky/pre-commit`, but it's not working.

`.husky/commit-msg` works.


typicode:
thank you!

PR #836, opened 2020-12-31
upgrade find versions to 4.0.0

k0nserv:
`find-versions` 4.0.0 depends on `^3.1.2` of `semver-regex`. Previous
versions of `semver-regex` contained a [vulnernability][0]. This was
unlikely to affect `husky` because of how `find-versions` is used, but
the only breaking change in `find-versions` is increasing the minimum
Node.js verison to 10, which is the same as Husky's minimum.

[0]: https://app.snyk.io/vuln/SNYK-JS-SEMVERREGEX-1047770

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3028,11 +3028,11 @@
       }
     },
     "find-versions": {
-      "version": "3.2.0",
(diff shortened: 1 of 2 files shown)


k0nserv:
Closing due to incorrect base branch

PR #870, opened 2021-02-14
style(shell): add trailing newlines

valtlai:
Adds trailing newlines to files generated by the `install` and `add` commands.

PR #876, opened 2021-02-16
Update .gitignore to include newline

cristobal:
Add newline to satisfy source controls systems such as Github and other automated build tools that add newline at the end of file. Otherwise we end up having people commiting this file to the repo when there's no real change.

So when i run npm install and having the `husky install` as part of the postinstall script it will overwrite the existing husky files even if they were already installed, or if a new collaborator comes to the team and runs npm install they will now have their `.husky/.gitignore` overwritten e.g.:

Screenshot 2021-02-16 at 22 41 43


cristobal:
Not sure what made the CI fail here 🤷🏽‍♂️

cristobal:
Duplicate of #870

typicode:
Hi @cristobal,

How could I reproduce the issue locally?

cristobal:
@typicode no worries this is solved in #870, just what @valtlai mentioned:
> It’s a very common practice to have a newline at the end of a file. I think Husky should align with that.

PR #886, opened 2021-02-25
docs: use npx vs. npm exec

blimmer:
`npm exec` is only available on NPM 7+. Using `npx` is cross-compatible with npm 6 and 7.


typicode:
Thank you :)

blimmer:
No problem - thank you for Husky!!

PR #903, opened 2021-03-18
fix: skipping installation/uninstallation inside node_modules (husky@4)

NikitaSikalov:
We use `husky@4` in our company as a dependency for our own "husky" package for configuring hooks in our own vcs system. This vcs system is similar to git but has some important features for our company. I moved the logic of skipping installation/uninstallation from `installer/index.ts` -> `installer/bin.ts` because when this package is installed/uninstalled as a transitive dependency it confuses users with errors messages. 
Also, I think these changes could be useful for users with other vcs systems and who also for some reason use this package.

### DIFF
--- a/src/installer/__tests__/index.ts
+++ b/src/installer/__tests__/index.ts
@@ -36,17 +36,8 @@ function install({
   })
 }
 
-function uninstall({
-  gitCommonDir = '.git',
-  userPkgDir = '.',
-}: {
-  gitCommonDir?: string
-  userPkgDir?: string
-} = {}): void {
-  installer.uninstall({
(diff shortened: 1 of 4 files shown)


typicode:
Thanks for the PR. There seems to be a linting issue that prevents running test. Could you run `npm run fix` and push changes?

NikitaSikalov:
So, as you can see there are still some errors in CI. However, it seems that these erros also exist in `HEAD` of base branch. Should I also fix them?

typicode:
> However, it seems that these erros also exist in HEAD of base branch. Should I also fix them?
If they already exists, no need (not as part of this PR at least :).

PR #906, opened 2021-03-19
Add required uninstallation step

laura-barluzzi:
After uninstalling husky v5 with `npm uninstall`, husky v4 didn't work. As it turned out, the `core.hooksPath` was still set to `.husky` so I had to unset the git config value for a clean uninstallation of husky v5.


typicode:
Interesting thanks for the report and PR. I would have expected husky's `uninstall` script to do it automatically. I'll do some more testing.

PR #911, opened 2021-03-24
Delete LICENSE-MIT

drpicox:
This file does not conform `package.json` `license` field.

> "to deal in the Software without restriction"


voxpelli:
Been solved in `master`

drpicox:
I did that because there are automatic processes that "got confused" about the licensing because the MIT license was present. Thanks!

PR #932, opened 2021-04-02
refactor(husky): remove redundant exit on husky.sh

DarkSuniuM:
Since the `exitCode` is defined on line 22, It is possible to directly exit on line 28 and keep the `exitCode`, Line 24 only checks if `exitCode` is not 0 and echoes the error.


typicode:
Good catch, thanks :+1:

PR #951, opened 2021-04-16
Don't create .gitignore

### LINKED ISSUE #944: No such file or directory
The first time pre-commit hooks after husky installation worked fine, but later on, out of nowhere, I am started seeing the below issue.

`.husky/pre-commit: line 3: .husky/_/husky.sh: No such file or directory`

husky version ^3.1.0

DoctorDerek:
This would close #944.

Current behavior:

- Husky creates a .gitignore file that ignores all the scripts in `.husky/_`
- Anyone creating a new git branch has to run `npm run prepare`... every single time
- Anyone using custom scripts (like `common.sh` [recommended in the docs](https://typicode.github.io/husky/#/?id=yarn-on-windows)) has to copy them manually

Desired behavior:

- Husky doesn't create a .gitignore file at all
- Husky reduces maintainability instead of increasing it
- The docs for every project using husky don't have to be updated to say "Please run `npm run prepare` if you ever make a new git branch... or clone the git repo without npm install... or ever want to use husky"

### DIFF
--- a/packages/husky/src/index.ts
+++ b/packages/husky/src/index.ts
@@ -31,7 +31,7 @@ export function install(dir = '.husky'): void {
     fs.mkdirSync(path.join(dir, '_'), { recursive: true })
 
     // Create .husky/.gitignore
-    fs.writeFileSync(path.join(dir, '.gitignore'), '_\n')
+    // fs.writeFileSync(path.join(dir, '.gitignore'), '_\n')
 
     // Copy husky.sh to .husky/_/husky.sh
     fs.copyFileSync(


DoctorDerek:
I understand that the current workflow is supposed to be `npm install` hitting the `prepare` script from `package.json` ... as alluded to by https://blog.typicode.com/husky-git-hooks-autoinstall/

But it's not the least bit clear why:
1. You have to `npm run prepare` in every new Git branch
2. You would have to manually handle custom scripts like `common.sh`
3. The docs seem to imply `npx husky-init` is the solution, not `npm run prepare`

DoctorDerek:
And, in fairness, if someone has the code locally and has already run `npm install` -- they should carry over the `.gitignore`d husky script file into the new branch.

But if that file were checked into the git repository, because this `.gitignore` file didn't exist, then it would be the same effect.

The current PR wouldn't even change `~/.husky/_husky.sh` from getting overwritten every time someone does `npm run script` because of `npm install` or otherwise.

--

The only rationale I can imagine for this particular solution as implemented in Husky v6 iis to increase support for multiple operating systems / package managers by allowing people to keep a folder on their desktop with their custom husky scripts, and then manually copy-paste those scripts over after doing `npm install`.

In that case, developers working on the same project with vastly different systems wouldn't have to manually edit their normal `.gitignore` file to include `.husky/_` ... but that seems like the odd use case, and one easy to add to the docs.

Currently, it's kind of the assumption... or maybe it has something to do with CI/CD deplloyments.

PR #959, opened 2021-05-02
docs: add git init troubleshoot

smhmd:
I was setting up a project from the remnants of another, and I only ran `git init` after setting up everything. It goes without a saying that the hooks didn't work.

If facing this situation seems unlikely, feel free to close the PR.


bfaulk96 (approved):

yordis:
ping

yordis:
ping

devinrhode2:
I think a better way to fix this would be anunder-the-hood check to see if the current directory is a git repo, and if not, fail with an error message. I do think this is rare enough that it's better not to clutter the docs with this info

smhmd:
@devinrhode2 Agreed!

PR #988, opened 2021-06-17
Updated README.md

nbelakovski:
Expanded -D into the long option, so that those who aren't as familiar with the ecosystem don't have to look it up.


typicode:
thank you :+1:

PR #989, opened 2021-06-18
removed duplicated code

fearphage:
The code removed is also the first line of `appendScript` which is called immediately afterwards. No need to do it *N* times. Once is enough.


typicode:
Thank you :+1:

Issue #1000, opened 2021-07-01
v7 Installer is no longer gracefully failing in CI context


cmdcarini:
**Context**
When running builds w/ Husky v7 in the CI, it seems that the Husky install is no longer failing gracefully, and is skipping over the `git rev-parse` [command](https://github.com/typicode/husky/blob/d47ffe1b6d80eaab3e1706d8a22562c737beb888/src/index.ts#L14) that usually halts the install process right away. 

Instead the install is now failing on [line 27 of the installer](https://github.com/typicode/husky/blob/d47ffe1b6d80eaab3e1706d8a22562c737beb888/src/index.ts#L27), that it can't find a .git directory.

typicode:
That would be a regression

typicode:
I did some tests.

It "works" like before except that the message is the one from git instead of a custom one.
But both in v6 or v7, the exit code is 0.

Instead of showing a custom "husky - not a Git repository, skipping hooks installation", I'm thinking that maybe showing Git error message is better. As you can find more info on the web with "fatal: not a git repository ..." rather than the husky one.

typicode:
That said, in CI context you can disable `prepare` script if you want:
https://typicode.github.io/husky/#/?id=disable-husky-in-cidocker

cmdcarini:
Hmm, ok. I'll have to take another look and see what's actually wrong with our use. The Docker use-case seems to be exactly what I'm looking for, I'll give that a shot. 

Thanks for the feedback and review!

typicode:
Thanks for the PR :)

PR #1001, opened 2021-07-01
fix: cover case where git rev-parse returns null

cmdcarini:
This resolves [this issue](https://github.com/typicode/husky/issues/1000) by covering for the case that `git rev-parse` returns null, which may be the case in a CI Docker image that does not bundle the git cli

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -11,7 +11,7 @@ const git = (args: string[]): cp.SpawnSyncReturns<Buffer> =>
 
 export function install(dir = '.husky'): void {
   // Ensure that we're inside a git repository
-  if (git(['rev-parse']).status) {
+  if (git(['rev-parse']).status !== 0) {
     return
   }
 


src/index.ts
```diff
@@ -11,7 +11,7 @@ const git = (args: string[]): cp.SpawnSyncReturns<Buffer> =>
 
 export function install(dir = '.husky'): void {
   // Ensure that we're inside a git repository
-  if (git(['rev-parse']).status) {
```
typicode, line 14: Except, if I'm missing something, it's equivalent?

```
$ node
Welcome to Node.js v16.2.0.
Type ".help" for more information.
> 0 ? "a" : "b"
'b'
> 1 ? "a" : "b"
'a'
>
```

typicode:
My bad, I didn't read correctly the title. `null` is returned if `git` command is not found.
Thanks for the PR!

PR #1005, opened 2021-07-07
fix: readme confusing condition to disable install in CI

Jolg42:
I think it's easier to read and the condition is now correct with the negation.


typicode:
Yes, thank you :+1:

PR #1023, opened 2021-08-10
Fix pre-commit hook in WebStorm under macOS

skyh:
I do not know why this works but this version of `husky.sh` now works well both in terminal and in WebStorm


skyh:
Found the reason: `set -e` with `&&` operator in `debug()`

typicode:
Interesting. Curious to know which shell was running.
Thanks for the fix :+1:

PR #1037, opened 2021-09-05
docs: fix typo on `docs/README.md`

shuuji3:
Hello, this PR fixes a tiny typo in `docs/README.md`.

I think the `--only` parameter of `npm ci` should be `--only=production` same as Line 254.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -251,14 +251,14 @@ There's no right or wrong way to disable husky in CI/Docker context and is highl
 If you want to prevent husky from installing completely
 
 ```shell
-npm ci --only=production --ignore-scripts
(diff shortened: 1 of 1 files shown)


typicode:
hello, good catch :+1: 

Actually it seems to be deprecated (https://docs.npmjs.com/cli/v7/using-npm/config#only) in favor of `--omit=dev`.

Could you update this PR to replace `--only` with `--omit` in the docs?

shuuji3:
@typicode You are right.🙂 I've updated the PR to use `--omit`.

typicode:
Thank you :)

shuuji3:
Thanks!

loversun08:
@blimmer

Issue #1034, opened 2021-09-01
Docs improvement suggestion 


jcollum:
> Edit package.json > prepare script and run it once

But I'm not actually editing the package.json, I'm using this script to add a script to it. It threw me off for a second when I read it. Plus, the `>` made me think I was doing redirection somehow. 

So I'd change that to 

> Run these two lines in a terminal: 
> 
> ```
> npm set-script prepare "husky install"  # adds a prepare script to package json
> npm run prepare # runs the script we just added 
> ```

typicode:
Thanks for the feedback @jcollum !

 Agree, would you like to make a PR with the above changes (would need to be changed in 3 places)?
Also for the `>`, you're right, maybe it could be rewritten differently?

gefymarcos:
I'd like to make a PR to resolve this, perhups changing ```package.json > prepare``` to ```prepare script to package.json``` make it clearer.

PR #1056, opened 2021-10-01
Update README.md

gefymarcos:
Docs improvement to make script preparation easier to understand, according to issue #1034


devinrhode2:
I've never seen a .github/README.md file before.. but that is unrelated to your change.

PR #1083, opened 2021-12-10
ci: Run on PRs, add linting, test on NPM 8

nschonni:
Could split these up, but figured they are easier to review together.

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -3,7 +3,9 @@
 
 name: Node.js CI
 
-on: [push]
+on:
+  push:
+  pull_request:
 
 jobs:
   build:
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -8,9 +8,21 @@ on:
   pull_request:
 
 jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


nschonni:
Sample run can be seen here https://github.com/nschonni/husky/actions/runs/1564351263

.github/workflows/node.js.yml
```diff
@@ -3,17 +3,34 @@
...
+      - name: Use Node.js
+        uses: actions/setup-node@v2
+        with:
+          cache: npm
+      - run: npm ci --ignore-scripts
+      - run: npm run lint
+
   build:
-    name: Test on node ${{ matrix.node-version }} and ${{ matrix.os }}
+    name: ${{ matrix.os }} with NodeJS ${{ matrix.node-version }} and NPM ${{ matrix.npm }}
```
XhmikosR, line 23: Nitpicking but `npm` is supposed to be spelled lowercase :)

typicode:
Thanks for the PR 👍 . 

I think for now I'd prefer to only test for the current latest version of npm. Mainly to keep tests fast and because husky code is pretty straightforward. There were also no particular issues with different versions of npm (except one on Windows that is fixed).

If you want you can make a PR just bumping npm from 7 to 8 in https://github.com/typicode/husky/blob/main/.github/workflows/node.js.yml

PR #1051, opened 2021-09-24
Use /usr/bin/env sh instead of direct path of sh

skhaz:
According with the best practices of shell scripting, the right way to invoke the _shebang_ is `#!/usr/bin/env shell`

Reference: https://www.cyberciti.biz/tips/finding-bash-perl-python-portably-using-env.html

PR #1089, opened 2022-01-08
Do not depend on the exact location of the 'sh' interpreter

utdemir:
Some Unix systems do not have sh at the specified location: `/bin/sh`. In my instance, I'm using [NixOS](https://nixos.org/), which also lacks it.

Using `/usr/bin/env sh` instead uses the `sh` executable from users `$PATH`, which should be more portable and also it makes it possible for the user to choose which 'sh' to use.

See: https://en.wikipedia.org/wiki/Shebang_(Unix)#Portability

### DIFF
--- a/.husky/commit-msg
+++ b/.husky/commit-msg
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env sh
 . "$(dirname "$0")/_/husky.sh"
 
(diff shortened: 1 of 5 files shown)


typicode:
Thanks for the PR 👍 
_Closing as https://github.com/typicode/husky/pull/1051 was merged_

PR #1092, opened 2022-01-12
add `husky -` prefix to logged global error messages

joshbalfour:
Hiya,

Spent some time today debugging an issue that I eventually traced back to husky, would have been nice to have the error message identified as such.

Admit this is probably more a yarn issue as all yarn gave me to go on was:

```
➤ YN0000: │ root@workspace:. STDERR fatal: not a git repository (or any of the parent directories): .git
```

But would appreciate this workaround being added, as it also bring it in line with [your other logging](https://github.com/typicode/husky/blob/main/src/index.ts#L6).

Cheers!

### DIFF
--- a/src/bin.ts
+++ b/src/bin.ts
@@ -38,6 +38,6 @@ try {
   cmds[cmd] ? cmds[cmd]() : help(0)
 } catch (e) {
   // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
-  console.error(e.message)
+  console.error(`husky - ${e.message}`)
   process.exit(1)
 }


devinrhode2:
Yarn 2 logs are ducking trash

typicode:
Thank you for the PR. Right now I can't see any output with Yarn 3. `postinstall` doesn't seem to be running, not sure why.

Could you try on your side with Yarn 3?

devinrhode2:
I think this is a great small change.

While yarn logging is very colorful, it has some major issues around exposing what's happening wrt npm scripts. I can say that writing to a file may be a reliable way to gain some insight into what npm scripts are running.

However, this is wholy un-necessary, looking at this code change, we can see it's extremely minor change. It may be good to double check that "husky - husky - FOO ERROR" is not possible

We could import a shared const for this string prefix, but better than that would be to have a shared logging function which adds the prefix if the prefix isn't already added. That way a shared const is not necessary.

devinrhode2 (approved):

typicode:
Thank you @devinrhode2 @joshbalfour

devinrhode2:
@typicode sorry my reply seemingly doesn't answer your question

> Thank you for the PR. Right now I can't see any output with Yarn 3. postinstall doesn't seem to be running, not sure why.
> 
> Could you try on your side with Yarn 3?

It does seem that postinstall is supported by yarn 3 here: https://yarnpkg.com/advanced/lifecycle-scripts

However, as I mentioned, you may not see your console.logs, but the script may still running. Outputting to file is one option, there are probably better strategies, maybe even a config option

PR #1095, opened 2022-01-23
feat: add dog emoji as favicon

jamiehaywood:
Thought it would look nice to add a dog emoji as a favicon for the docs site.
[It has ok browser support](https://caniuse.com/link-icon-svg), but I can generate and add standard `.ico` and `pngs` for wider support if it's something that you would like.


typicode:
Agree 😄

PR #1098, opened 2022-01-29
Custom logger feature

gautaz:
Hello @typicode,

I am using husky as part of a scaffolding project and needed to pass husky the logger used from the outer scope.

I do not know if this feature is something you might want to add in husky but here it is.
I am quite new to typescript so I might have made some mistakes when specifying the logger interface.
This interface defines 3 levels of log, one of which is currently unused (warn), I can remove it if you feel like it will never be necessary or if you want to add it later.

I have also added a test case and extended the documentation a bit.
I am open to any advice/modification to make this go upstream if this is something you are ok with.

Gautaz

**EDIT:**

I have replaced the `logger` argument by `customOptions` for the `configure` method, it seems more relevant and `CustomOptions` can later be extended with additional features.

I have also removed the `warn` logger as it was not used (it can be added later anyway if needed).

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -282,6 +282,18 @@ if (!isCi) {
 }
 ```
 
+You can also use your own logging functions if needed:
+
+```js
+// prepare.js
+const { configure } = require('husky)
+const husky = configure({
+  log: (msg) => console.log(msg),
+  error: (msg) => console.error(`Something went bad: ${msg}`)
+})
+husky.install()
+```
+
 Or make `prepare` script fail silently if husky is not installed:
 
 ```json
--- a/docs/README.md
+++ b/docs/README.md
@@ -286,7 +286,7 @@ You can also use your own logging functions if needed:
 
 ```js
 // prepare.js
-const { configure } = require('husky)
+const { configure } = require('husky')
 const husky = configure({
   log: (msg) => console.log(msg),
   error: (msg) => console.error(`Something went bad: ${msg}`)
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,7 +21,7 @@
         "typescript": "^4.6.3"
       },
       "engines": {
-        "node": ">=12"
+        "node": ">=14"
       },
       "funding": {
         "url": "https://github.com/sponsors/typicode"
--- a/src/index.ts
+++ b/src/index.ts
@@ -3,8 +3,10 @@ import fs = require('fs')
 import p = require('path')
 
 interface Options {
-  log: (msg: string) => void
-  error: (msg: string) => void
+  logger: {
+    log: (msg: string) => void
+    error: (msg: string) => void
+  }
 }
 
 type CustomOptions = Partial<Options>
@@ -18,8 +20,10 @@ interface Husky {
 
 // Default options
 const defaultOptions: Options = {
-  log: (msg: string): void => console.log(`husky - ${msg}`),
-  error: (msg: string): void => console.error(`husky - ${msg}`),
+  logger: {
+    log: (msg: string): void => console.log(`husky - ${msg}`),
+    error: (msg: string): void => console.error(`husky - ${msg}`),
+  },
 }
 
 // Git command
@@ -29,11 +33,11 @@ const git = (args: string[]): cp.SpawnSyncReturns<Buffer> =>
 const defaultHusky = configure()
 
 export function configure(customOptions: CustomOptions = {}): Husky {
-  const options: Options = { ...defaultOptions, ...customOptions }
+  const { logger }: Options = { ...defaultOptions, ...customOptions }
   return {
     install(dir = '.husky'): void {
       if (process.env.HUSKY === '0') {
-        options.log('HUSKY env variable is set to 0, skipping install')
+        logger.log('HUSKY env variable is set to 0, skipping install')
         return
       }
 
@@ -76,11 +80,11 @@ export function configure(customOptions: CustomOptions = {}): Husky {
           throw error
         }
       } catch (e) {
-        options.error('Git hooks failed to install')
+        logger.error('Git hooks failed to install')
         throw e
       }
 
-      options.log('Git hooks installed')
+      logger.log('Git hooks installed')
     },
 
     set(file: string, cmd: string): void {
@@ -101,13 +105,13 @@ ${cmd}
         { mode: 0o0755 },
       )
 
-      options.log(`created ${file}`)
+      logger.log(`created ${file}`)
     },
 
     add(file: string, cmd: string): void {
       if (fs.existsSync(file)) {
         fs.appendFileSync(file, `${cmd}\n`)
-        options.log(`updated ${file}`)
+        logger.log(`updated ${file}`)
       } else {
         this.set(file, cmd)
       }
--- a/src/index.ts
+++ b/src/index.ts
@@ -2,14 +2,14 @@ import cp = require('child_process')
 import fs = require('fs')
 import p = require('path')
 
-interface Options {
-  logger: {
-    log: (msg: string) => void
-    error: (msg: string) => void
-  }
+interface Logger {
+  log: (msg: string) => void
+  error: (msg: string) => void
 }
 
-type CustomOptions = Partial<Options>
+interface Options {
+  logger?: Partial<Logger>
+}
 
 interface Husky {
   install: (dir: string) => void
@@ -18,12 +18,9 @@ interface Husky {
   uninstall: () => void
 }
 
-// Default options
-const defaultOptions: Options = {
-  logger: {
-    log: (msg: string): void => console.log(`husky - ${msg}`),
-    error: (msg: string): void => console.error(`husky - ${msg}`),
-  },
+const defaultLogger: Logger = {
+  log: (msg: string): void => console.log(`husky - ${msg}`),
+  error: (msg: string): void => console.error(`husky - ${msg}`),
 }
 
 // Git command
@@ -32,8 +29,12 @@ const git = (args: string[]): cp.SpawnSyncReturns<Buffer> =>
 
 const defaultHusky = configure()
 
-export function configure(customOptions: CustomOptions = {}): Husky {
-  const { logger }: Options = { ...defaultOptions, ...customOptions }
+export function configure(customOptions: Options = {}): Husky {
+  const logger: Logger = {
+    log: customOptions?.logger?.log || defaultLogger.log,
+    error: customOptions?.logger?.log || defaultLogger.error,
+  }
+
   return {
     install(dir = '.husky'): void {
       if (process.env.HUSKY === '0') {
@@ -122,6 +123,7 @@ ${cmd}
     },
   }
 }
+
 export const install = defaultHusky.install.bind(defaultHusky)
 export const set = defaultHusky.set.bind(defaultHusky)
 export const add = defaultHusky.add.bind(defaultHusky)
--- a/src/index.ts
+++ b/src/index.ts
@@ -2,92 +2,126 @@ import cp = require('child_process')
 import fs = require('fs')
 import p = require('path')
 
-// Logger
-const l = (msg: string): void => console.log(`husky - ${msg}`)
+interface Options {
+  log: (msg: string) => void
+  error: (msg: string) => void
+}
+
+interface CustomOptions {
+  log?: (msg: string) => void
+  error?: (msg: string) => void
+}
+
+interface Husky {
+  install: (dir: string) => void
+  set: (file: string, cmd: string) => void
+  add: (file: string, cmd: string) => void
+  uninstall: () => void
+}
+
+// Default options
+const defaultOptions: Options = {
+  log: (msg: string): void => console.log(`husky - ${msg}`),
+  error: (msg: string): void => console.error(`husky - ${msg}`),
+}
 
 // Git command
 const git = (args: string[]): cp.SpawnSyncReturns<Buffer> =>
   cp.spawnSync('git', args, { stdio: 'inherit' })
 
-export function install(dir = '.husky'): void {
-  if (process.env.HUSKY === '0') {
-    l('HUSKY env variable is set to 0, skipping install')
-    return
-  }
+const defaultHusky = configure()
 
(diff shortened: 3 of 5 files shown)


devinrhode2:
Foff stale bot

gautaz:
@devinrhode2 Thanks for taking care of the stale bot :-).

I took the opportunity to take into account the latest commits on the main branch.

docs/README.md
```diff
@@ -279,6 +279,18 @@ const isCi = process.env.CI !== undefined
 if (!isCi) {
   require('husky').install()
 }
+
+
+You can also use your own logging functions if needed:
```
devinrhode2, line 284: Should this be outside if the code formatted block?

src/index.ts
```diff
@@ -2,87 +2,121 @@ import cp = require('child_process')
...
-    throw new Error(`.git can't be found (see ${url})`)
-  }
-
-  try {
-    // Create .husky/_
-    fs.mkdirSync(p.join(dir, '_'), { recursive: true })
-
-    // Create .husky/_/.gitignore
-    fs.writeFileSync(p.join(dir, '_/.gitignore'), '*')
+interface Options {
```
devinrhode2, line 5: In your 2nd commit you switched from the name Logger to Options

Having recently worked with pino, I think Logger is a better name (please mind that I'm just casually reviewing here)

devinrhode2:
> @devinrhode2 Thanks for taking care of the stale bot :-).
> 
> 
> 
> I took the opportunity to take into account the latest commits on the main branch.

Is there any good way I can view the diff from before+after your force push?

docs/README.md
gautaz: You're right, I messed up while resolving conflicts.
I'll fix that right away.

I miss so much the unit tests for documentation ;-).

src/index.ts
gautaz: That was my intention at first, the initial draft of the PR was in fact providing a way to pass a custom logger.

Then I realized that if one wanted later to add other options apart from logging functions, the `configure` function would take multiple arguments: `configure(logger: Logger = defaultLogger, customOptions: CustomOptions = {})`.

I found this rather awkward to separate optional custom logging functions from other potential future options.

I know it's a rather opinionated choice, perhaps having multiple opinions from different people on this matter might help?

docs/README.md
gautaz: fixed.

gautaz:
@devinrhode2 
> Is there any good way I can view the diff from before+after your force push?

AFAIK there is no way to do that because a force push basically rewrites the branch history.
The reason why I am rewriting the branch history is to avoid multiple merge commits until the branch is merged on the trunk.
This avoids loosing the initial commits (the ones that matter in the PR) in a flow of successive merge commits.

It eases the review for a new reviewer but forces persons already reviewing to review the whole PR each time...
I guess the choice has to be made depending on the size of the PR.
As for this PR, I do not know if you consider it to be a small or a big one 😃.

src/index.ts
devinrhode2: Upon re reading this I agree "Options" is a better name

gautaz:
Hello @typicode,

I don't mean to be rude but would you please mind having a look at this PR and give me some feedback?

If the current implementation does not match your standards, would you please tell me what would be the right way to do it?

Thanks for your time,

Gautaz

devinrhode2:
I was cleaning up diff, and not seeing anything alarming, but saw something very interesting:
CleanShot 2022-06-06 at 18 47 31

Good commit history reveals these little bits and pieces making PRs easier to review. After merge, the history is still quite useful for anyone to understand what the actual changes were :)

The bare minimum we need from the tooling ecosystem is for the `git diff` to at least be able to ignore whitespace/indentation changes!

gautaz:
Hello @devinrhode2 , thanks for the review :smile:.

Yes, the default github diff does not ignore whitespaces, you can change that easily : https://stackoverflow.com/questions/37007300/how-to-ignore-whitespace-in-github-when-comparing

Applying this renders the following (which shares similarities with your diff but takes into account all my modifications):
image

As for the `this.set`, it is necessary to call the `set` method on the current husky instance produced by the `configure` function.
The drawback is that I had to bind all original exported husky functions as shown at the bottom of my screenshot.

I am open to any suggestion on this matter.

devinrhode2:
You gave me an idea!

I'm not sure it's truly better, but you could invert things

Instead of housing all the code inside configure, you could turn each method into createInstall, createAdd, etc, 

But I think that might produce more code, overall

I think it's good as is!

@typicode note that guataz did change package lock to update all minor versions

Q: I generally like to see accurate versions inside of package.json, but since this is a classic npm/node module... I'm not sure what would be best

My commit history explicitly calls out this package-lock update, which I think would be very good to separate, from what I can tell the minor version bump is completely orthogonal and unrelated, making it a separate commit helps with that. 

@guataz I'm just personally curious, what npm command did you run to perform this update?

I wonder if we should only let dependabot update dependencies?

gautaz:
@devinrhode2 You're right, it's best if the PR does not play with the dependencies versions, it is unrelated to the PR original intent.
I will remove this from the PR.

As for the command I used to update the `package-lock.json`, I may have done something a bit drastic at this time like `rm package-lock.json && npm i` while handling some conflicts during a rebase (this feels stupid now that I am looking back to it as I should not have modified the lock file at all).

gautaz:
Done.

The `package-lock.json` file is still currently modified by `npm i` for just on line:
image

Which is because its node requirement is not in sync with the one detailed in `package.json`.
Is that an issue ?

docs/README.md
```diff
@@ -282,6 +282,18 @@ if (!isCi) {
 }
 ```
 
+You can also use your own logging functions if needed:
+
+```js
+// prepare.js
+const { configure } = require('husky)
```
devinrhode2, line 289: I see your point about combined unit tests/documentation - there is a missing single quote in `require('husky)`

package-lock.json
```diff
@@ -21,7 +21,7 @@
         "typescript": "^4.6.3"
       },
       "engines": {
-        "node": ">=12"
```
devinrhode2, line 24: +1

devinrhode2:
Oh man :(

I had bunch of pending comments that were not submitted
only like 1-2 things to tweak

test/8_custom_logger.sh
```diff
@@ -0,0 +1,29 @@
+. "$(dirname "$0")/functions.sh"
```
devinrhode2, line 1: you probably know more bash/zsh/sh than me...
but the quotes around `"$0"` don't seem right
```sh
. "$(dirname "$0")/functions.sh"
```
wouldn't this produce a syntax error?
```sh
"$(dirname "
```

docs/README.md
devinrhode2: LGTM

src/index.ts
```diff
@@ -2,92 +2,126 @@ import cp = require('child_process')
...
+  log: (msg: string) => void
+  error: (msg: string) => void
+}
 
-    // Copy husky.sh to .husky/_/husky.sh
-    fs.copyFileSync(p.join(__dirname, '../husky.sh'), p.join(dir, '_/husky.sh'))
+interface CustomOptions {
+  log?: (msg: string) => void
+  error?: (msg: string) => void
+}
```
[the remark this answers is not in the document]
devinrhode2: TODO ^

src/index.ts
```diff
@@ -2,92 +2,126 @@ import cp = require('child_process')
...
 import p = require('path')
 
-// Logger
-const l = (msg: string): void => console.log(`husky - ${msg}`)
-
-// Git command
-const git = (args: string[]): cp.SpawnSyncReturns<Buffer> =>
-  cp.spawnSync('git', args, { stdio: 'inherit' })
-
-export function install(dir = '.husky'): void {
```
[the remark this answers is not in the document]
devinrhode2: rebasing is not very meaningful anymore
LGTM

devinrhode2:
> Done.
> 
> The `package-lock.json` file is still currently modified by `npm i` for just on line: image
> 
> Which is because its node requirement is not in sync with the one detailed in `package.json`. Is that an issue ?

It's good to get it updated and in sync, that's for sure :)

devinrhode2:
I have 4 "TODO" comments, 2 "optional", 2 seem more meaningful

gautaz:
> I have 4 "TODO" comments, 2 "optional", 2 seem more meaningful

Thanks a lot @devinrhode2 :smile:, I'll look into them ASAP.

test/8_custom_logger.sh
gautaz: It should be valid because the inner `"$0"` is encased in a `$(...)` command substitution construct.

The [bash documentation](https://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html#Command-Substitution) tells that:
> When using the $(command) form, all characters between the parentheses make up the command; none are treated specially.

The only issue occurs when using backquotes in backquotes:
> Command substitutions may be nested. To nest when using the backquoted form, escape the inner backquotes with backslashes.

The [zsh documentation](https://zsh.sourceforge.io/Doc/Release/Expansion.html#Command-Substitution) basically tells the same thing.

For sh POSIX shells, the [documentation](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_03) basically boils down to this: the substituted command is executed in a subshell as is.

I guess that should cover the potential issue :relieved:.

src/index.ts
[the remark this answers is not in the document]
gautaz: Just fixed this one.
I don't think the performance impact will ever be noticeable, so the `Partial<Options>` seems ok.

gautaz:
Hello @devinrhode2,

A little recap of the latest modifications:
- the missing backquote in documentation has been fixed
- I followed your recommendation to use `Partial` for the `CustomOptions` declaration
- I ensured that there is no shell issue with `. "$(dirname "$0")/functions.sh"`

Your latest comment mentioned a fourth TODO but I didn't found it... :disappointed:

src/index.ts
```diff
@@ -2,92 +2,126 @@ import cp = require('child_process')
...
+const git = (args: string[]): cp.SpawnSyncReturns<Buffer> =>
+  cp.spawnSync('git', args, { stdio: 'inherit' })
 
-  fs.writeFileSync(
-    file,
-    `#!/usr/bin/env sh
+const defaultHusky = configure()
+
+export function configure(customOptions: CustomOptions = {}): Husky {
+  const options: Options = { ...defaultOptions, ...customOptions }
```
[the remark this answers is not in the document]
devinrhode2: Here's the 4th one :)

Sorry I am not that experienced with github's review flow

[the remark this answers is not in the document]
gautaz: Me neither :smile:.

I agree that the current naming is quite odd and your proposal is more readable.
Let's change that.

devinrhode2:
Having written this yourself @gautaz, what do you think about the warn function again?

I was thinking that if someone wants to pass in a warn function, it shouldn't cause a type error?

I think I am bike shedding at this point, pr can be merged once guataz is ready

src/index.ts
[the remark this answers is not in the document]
gautaz: I've just embedded the log functions in a `logger` section within `Options` which is clearly more readable.

The drawback from a TypeScript point of view is that one has to define both `log` and `error` functions if the `logger` is overridden (I have checked that TypeScript will raise an error if `logger` is partially overridden).
This seems ok as it would be odd for someone to only partially redefine the logger.

What I am more concerned about is people using husky in pure javascript.
As they do not have the TypeScript safety net, they might forgot to override all `logger` functions and this will fail badly at runtime.

So I am wondering if I should not make all the logger functions optional in order to avoid this situation...

gautaz:
> Having written this yourself @gautaz, what do you think about the warn function again?
> I was thinking that if someone wants to pass in a warn function, it shouldn't cause a type error?
> I think I am bike shedding at this point, pr can be merged once guataz is ready

@devinrhode2 I just have a doubt about your intent, do you mean that we should forbid someone to pass functions that would remain unused or do you mean that we should add a `warn` function again? (I am not an english native so I might misunderstood things from time to time)

gautaz:
Hello again @devinrhode2, I'm not sure you did see my latest comment regarding the warn function.
If it's not such a big deal, I guess the PR is ready for merging unless I have forgotten anything else to fix :sweat_smile:.

devinrhode2:
Sorry dropped the ball 

What I was trying to say, is that maybe we add this to the type interface:
warn?: (...args: any[]) => void

src/index.ts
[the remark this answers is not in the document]
gautaz: > What I am more concerned about is people using husky in pure javascript.
> As they do not have the TypeScript safety net, they might forgot to override all logger functions and this will fail badly at runtime.

I just refactored the logger initialization in order to avoid this type of issue.

gautaz:
> What I was trying to say, is that maybe we add this to the type interface: warn?: (...args: any[]) => void

I could have added this in my latest commit but it felt a bit too much as the function is not used for now.
Its ready for merging :-).

devinrhode2:
Agreed

@typicode this is ready to merge imo

devinrhode2:
nooooo stalebot!

gautaz:
> nooooo stalebot!

Thanks @devinrhode2 :thumbsup:!

@typicode, enjoy your vacations!

gautaz:
Hello @typicode, I can resolve conflicts once again but I cannot do this forever without knowing if this will ever be merged.

Can you please at least state your position regarding this PR?

typicode:
Truly sorry and thank you for the PR, I will review it for next major release (planned for Node EOL). I'll take care of the potential conflicts if it's merged, you can leave it as it is.

typicode:
Thanks again for the PR, after long consideration I'd rather not make husky's code more complex. Especially since it can be achieved without any change in the API.

If really needed (though I think husky shouldn't be silent since it's making changes to user's Git config), it's possible this way with the current version:

```js
const husky = require('husky')

console.log = () => {}
try { husky.install() } catch {}
```

PR #1104, opened 2022-02-16
Actually set 'husky_skip_init' as readonly in `./husky.sh`

hyperupcall:
Previously, the `husky.sh` script would export the empty 'readonly' variable, along with 'husky_skip_init' which has a value of 1.

Now, the script sets 'husky_skip_init' to 1, and properly gives is the 'readonly' and 'exported' attributes.

### DIFF
--- a/husky.sh
+++ b/husky.sh
@@ -19,7 +19,8 @@ if [ -z "$husky_skip_init" ]; then
     . ~/.huskyrc
   fi
 
-  export readonly husky_skip_init=1
+  readonly husky_skip_init=1
+  export husky_skip_init
   sh -e "$0" "$@"
   exitCode="$?"
 


typicode:
I may be missing something but I can't seem to reproduce the issue. I tried in bash:

```
~ 
❯ export readonly husky_skip_init=1

~ 
❯ echo $husky_skip_init
1
```

hyperupcall:
In both cases, the value of `husky_skip_init` is `1`, the difference is whether the variable is marked as readonly

Before the changes, a variable called `readonly` was spuriously created

```sh
$ export readonly husky_skip_init=1

$ declare -p readonly
declare -x readonly

$ declare -p husky_skip_init
declare -x husky_skip_init="1"
```

In other words, not readonly

```sh
$ export readonly husky_skip_init=1
$ husky_skip_init=2
$ # (no error)
```

It didn't seem to match up semantically, as in [other cases](https://github.com/typicode/husky/blob/2e4da5ce986a917bc73ff08d0f85d58b58fdb6a4/husky.sh#L9) variables were properly marked as `readonly`

So I changed the code so that `husky_skip_init` was actually readonly, since it seemed like that was the original intent. After the changes, we get

```sh
$ readonly husky_skip_init=1
$ export husky_skip_init

$ declare -p readonly
bash: declare: readonly: not found

$ declare -p husky_skip_init
declare -rx husky_skip_init="1"
```

As you can see, as a result, the `husky_skip_init` variable is marked as readonly (`-r`) and a variable called 'readonly' isn't spuriously created

```sh
$ readonly husky_skip_init=1
$ export husky_skip_init
$ husky_skip_init=2
bash: husky_skip_init: readonly variable
```

It doesn't make a huge difference in terms of behavior since the variable isn't used much or modified, but I'm sure for people not (or less) familiar with shell, the code is less confusing

typicode:
I see, good catch :+1: 
Thanks for the very detailed steps.

PR #1111, opened 2022-03-02
Directly re-execute the current script

### LINKED ISSUE #971: Support arbitrary shells for hook scripts
Husky seems to run all hook scripts using `sh`, regardless of the shebang present in the hook. For example, with a pre-commit hook containing a line like:

```bash
source "$(dirname "$0")/_/husky.sh"
```

we will get an error, because `source` is a Bash-specific command:

```shell-session
$ git commit
.husky/pre-commit: 4: source: not found
```

This occurs regardless of whether the shebang line is set as `#!/bin/sh` or `#!/usr/bin/env bash`, and seems to be because Husky invokes the currently-running hook script using sh in order to capture the exit value: https://github.com/typicode/husky/blob/a66271d21fdf798b291d009c213ed48737dae227/packages/husky/src/husky.sh#L20-L26

 A workaround for this is to use a Bourne-shell compatible `pre-commit` script, which then invokes bash explicitly with the target hook.

However, it would be nice if Husky respected the shebang of the hook script directly when executing it. This seems to be possible by changing husky.sh to use env to execute the command instead, for example:

```shell
  export readonly husky_skip_init=1
  env "$0" "$@"
  exitCode="$?"
```

However, this only works if the hook is also marked executable. Maybe we can use $SHELL instead to detect the currently-running shell, and exec the script again using that?

petekeen-cf:
The current `husky.sh` re-executes the current script with `sh -e`, completely ignoring whatever shebang line is in the script. This causes failures when scripts are expecting bash-specific features but `/bin/sh` is linked to something that is not bash (ex: `/bin/sh` is `dash` on Ubuntu).

This change makes it re-execute using `env` instead which will just run the script. This is safe because by definition [git hooks have to be executable](https://git-scm.com/docs/githooks/2.24.0):

> Hooks that don’t have the executable bit set are ignored.

Fixes #971.

### DIFF
--- a/husky.sh
+++ b/husky.sh
@@ -21,7 +21,7 @@ if [ -z "$husky_skip_init" ]; then
 
   readonly husky_skip_init=1
   export husky_skip_init
-  sh -e "$0" "$@"
+  env "$0" "$@"
   exitCode="$?"
 
   if [ $exitCode != 0 ]; then


typicode:
Hi,

Thanks for the PR :)
Changing to `env` changes husky current behavior, which is to exit on the first failing command.

For example:

```
#!/bin/sh
. "$(dirname "$0")/_/husky.sh"

foo
bar
baz
```

Will output with `env`:

```
.husky/pre-commit: line 4: foo: command not found
.husky/pre-commit: line 5: bar: command not found
.husky/pre-commit: line 6: baz: command not found
husky - pre-commit hook exited with code 127 (error)
```

vs with `sh -e`:

```
.husky/pre-commit: line 4: foo: command not found
husky - pre-commit hook exited with code 127 (error)
```

That said this PR was merged recently and makes use of `env` I don't know if that can help:
https://github.com/typicode/husky/pull/1051

It's not yet released though.

petekeen-cf:
I saw that PR but I don't think it solves this problem because those `#!` lines are entirely ignored on re-exec. They never get read because `sh -e` is executing the file as if it were a shell script. The kernel (or equivalent process in macOS) is what actually looks at the `#!` line and decides what to run. 

That said, I see where you're coming from with this behavior. I agree that it's nice to fail fast and to attempt to force that behavior on script authors. I'm just in an environment where we want our scripts to use non-POSIX `bash`isms like `set -o pipefail`, but re-executing with `sh` makes those fail on platforms like Ubuntu where `sh` is linked to `dash`.

I'll go ahead and close this out since it's not desired behavior.

typicode:
If husky's script gets in your way or your team, feel free to modify/commit it locally or simply remove it. The script may be useful in the case of an Open Source project where you don't know about other people environments, but if it's for a private project where you have more control, it can makes sense to not use it.

Alternatively, if you want to keep it and use bash, you can also do:

```sh
# .husky/pre-commit
#!/usr/bin/env sh
. "$(dirname "$0")/_/husky.sh"

bash your_bash_script
```

PR #1132, opened 2022-04-16
Force `basename`/`dirname` to treat `$0` as an argument

mataha:
`basename`/`dirname` will treat `$0` as an option if it starts with a `-`; e.g. for `-husky`:

```sh
dirname: invalid option -- 'h'
Try 'dirname --help' for more information.
```

Why did my directory start with a hyphen instead of a dot in the first place? I'd... rather not answer this, but it's related to my setup and not a bug in this project.


typicode:
Thank you!

PR #1147, opened 2022-05-13
Use number comparators when comparing exit codes

mataha:
This has exploded on me once before - when an equivalent of `exitCode` could be coerced to a number, but was an actual string.

Updated Node version in the lockfile as a by-product.

### DIFF
--- a/husky.sh
+++ b/husky.sh
@@ -24,13 +24,13 @@ if [ -z "$husky_skip_init" ]; then
   sh -e "$0" "$@"
   exitCode="$?"
 
-  if [ $exitCode != 0 ]; then
+  if [ "$exitCode" -ne 0 ]; then
     echo "husky - $hook_name hook exited with code $exitCode (error)"
   fi
 
-  if [ $exitCode = 127 ]; then
+  if [ "$exitCode" -eq 127 ]; then
     echo "husky - command not found in PATH=$PATH"
   fi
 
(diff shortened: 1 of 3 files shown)


devinrhode2:
Hmm.. I don't really know bash, but are there any downsides here? What kind of strings were being coerced to 0? Something like `"0: no more items in the array"`? Forgive me as I don't really know bash

mataha:
> Hmm.. I don't really know bash, but are there any downsides here? What kind of strings were being coerced to 0? Something like `"0: no more items in the array"`? Forgive me as I don't really know bash

The usual suspects were empty strings (e.g. `""`, `" "` etc.). Those scripts don't use Bash though - it's POSIX shell.

typicode:
Thanks for the PR, would you have some failing code to reproduce the issue?

mataha:
Sorry for taking so long to respond, couldn't reproduce this in my current shell - but it went roughly like this:

`code.bat`:
```batchfile
:: Shadow the exit code global variable
@set ERRORLEVEL=iddqd

:: Exit
@exit /b %ERRORLEVEL%
```

Shell (within WSL):
```sh
# Run the command starting from a Windows path
cmd.exe /c "pushd "$(wslpath -aw "$(pwd)")" && code.bat"

# Check the exit code
echo "$?"
```

PR #1156, opened 2022-06-11
feat: add env as a variable to give to husky to set the environment it is run in

Gentatsu:
This PR enables users to add as argument to the install script the environment the scripts run in. 
An example of this is to run in `bash` instead of `sh`. 

This PR does not include the scope of optional single args, however so it must be provided as a second argument to the dir.

Example usage `yarn run prepare .husky bash`  or apply directly in the `prepare` section of the `package.json`

This allows us to specify `zsh` as well as other shell environments.

### DIFF
--- a/src/bin.ts
+++ b/src/bin.ts
@@ -5,7 +5,7 @@ import h = require('./')
 // Show usage and exit with code
 function help(code: number) {
   console.log(`Usage:
-  husky install [dir] (default: .husky)
+  husky install [dir] (default: .husky) [env] (default sh)
   husky uninstall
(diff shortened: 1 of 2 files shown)


Gentatsu:
Bumpy

Gentatsu:
One more bump

typicode:
Thanks for the PR. However, not all shells are available. In particular on Windows.
So this feature could create issues for users. For better or worse, `sh` is one you can expect to work everywhere.

If you're working in a team where the available shells are known beforehand, you can instead have:

```
# husky/pre-commit
# ...
bash some_script
```

Issue #1166, opened 2022-06-29
Proposal: Unify the three README files


doublethefish:
# Problem:
There are three README files

* `/docs/README.md` - seen on the docs site.
* `/README.md` - seen only in local checkouts.
* `/.github/README.md` - used as the _default_ readme on github, and only obvious there.

This leads to:
1. drift and subtle differences between the docs.
2. making contributions harder and more confusing.

# Proposal:

Move `./docs/README.md` to `/README.md`, and delete the others. Obviously, there would have to be consideration given to generating the docs site.

# Not relevant to this issue:
> 
> **Troubleshoot**
> - [x] Before creating an issue, please check:
> https://typicode.github.io/husky/#/?id=troubleshoot
> 
> If you're migrating from husky 4, see:
> https://typicode.github.io/husky/#/?id=migrate-from-v4-to-v7
> 
> **Context**
> Please describe your issue and provide some context:
> - Terminal or GUI client (PowerShell, Git Bash, GitHub Desktop, ...)
> - If applicable, content of the failing hook
> - If possible, minimal steps to reproduce the issue
> 
> Thank you!

doublethefish:
This still needs consideration.

FilipMaslovaric:
So which one is the correct README.md to follow, and most up to date???

doublethefish:
> So which one is the correct README.md to follow, and most up to date???

Exactly! I’ve had PR #1167 open for a long time, but this project seems to be unsupported.

Basically, ‘./docs/README.md’ is mostly the most up to date.

typicode:
Thanks for rising this issue. READMEs have been updated and should reflect the same usage instructions. I've answered as to why there are three READMEs in the PR.

PR #1167, opened 2022-06-29
fix(#1166): docs/reduce READMEs

doublethefish:
Uses a single README.md addressing the problem defined in #1166.


typicode:
Thanks for the PR and your patience.

A bit of context regarding this choice:
- README on npm is small and straight to the point to avoid impacting package size too much
- README on GitHub is a little bigger with a link to the full doc so that people can still have enough to have a quick start with husky and go deeper if needed
- docs/README is the full docs

I agree that it can create sync problems. However, for now, I think it's OK. Also hopefully, `npm pkg set` should be stable now.

If anything, maybe later, I'd just have a link to the docs on npm.

doublethefish:
I’ve actually rewritten the docs so they’re easier to follow and read, in all contexts. Without this, very reasonable, PR, that work can’t continue.

Is this decision itself documented? I ask because it did make contribution harder.

The reasoning here is a very poor justification that causes more overhead for maintenance.

PR #1218, opened 2022-11-26
Adjust instructions for nvm workaround

kylemh:
I scoured a lot of GitHub issues and Stack Overflow problems. The difficulty in just using:

```sh
export NVM_DIR="$HOME/.nvm"
```

is that - by default - nvm.sh will run the default Node version which isn't necessarily the project's node version. This still won't save anybody manually bouncing between Node versions within different projects, but `.nvmrc` is a very common practice and this should assuage people who run Husky in a project using an `.nvmrc` file that differs from their NVM default.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -368,9 +368,16 @@ For example, for `nvm` that would be:
 
 ```shell
 # ~/.huskyrc
-# This loads nvm.sh and sets the correct PATH before running hook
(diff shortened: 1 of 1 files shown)


typicode:
Thanks for the PR. Out of curiosity, what's the output if you run `nvm use` in a directory with no `.nvmrc`?

kylemh:
Screenshot 2022-12-13 at 7 43 14 PM

It errors, which is why I only run `nvm use` if the file exists.

PR #1224, opened 2022-12-13
update docs to add message about restarting vs code

claycoleman:
VS Code can get into a weird state where husky `pre-commit`s will fail with `command not found` but the solution is to simply restart VS Code. see https://stackoverflow.com/a/73635509/5504797

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -373,7 +373,7 @@ export NVM_DIR="$HOME/.nvm"
 [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
 ```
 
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -373,6 +373,8 @@ export NVM_DIR="$HOME/.nvm"
 [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
 ```
 
+!> **for some apps (eg, VS Code), you can resolve this simply by restarting the app. try this before following any of these steps above!**
```
paescuj, line 376:
```suggestion
!> **For some apps (e.g., VS Code), you can resolve this simply by restarting the app. Try this before following any of these steps above!**
```

PR #1243, opened 2023-03-04
Make commit hooks created on Windows executable on Linux

### LINKED ISSUE #1177: hook was ignored because it's not set as executable
**Context**
I get following error on Mac/Linux machines:
```
hint: The '.husky/pre-commit' hook was ignored because it's not set as executable.
hint: You can disable this warning with git config advice.ignoredHook false.
```

**STR**
I added husky 7 on Windows according to the recommendations.
Then I committed all changes to git and updated on other machine (Mac/Linux).
I ran yarn update, which caused "prepare" script to run (husky install).
then I tried to change something and commit, which showed me above errors and didn't execute hooks.

**Expected behavior:**
Husky should somehow prepare all related files to automatically work on any platform without additional manual steps, like chmod

Thank you!

nathanpovo:
Fixes #1177

Solves the issue of a commit hook not being executable on Linux systems when the commit hook was created on a Windows system.

Previously, when a new commit hook was created on a Windows system, the executable bit (which is required by Git on Linux systems to run a commit hook) was not added to the newly created commit hook file because Windows does not keep track of the executable bit of a file. This resulted in commit hooks not working as intended when the repository was cloned onto a Linux system.

To fix this issue, the executable bit will now be added to the file when creating a new commit hook on a Windows system by using the Git command `git update-index --chmod=+x .\path\to\file`. This solution is a bit of a hack since it requires the user to manually commit the file to actually save the executable bit with the file. However, since Windows does not keep track of the executable bit, it is not possible to fix this issue without performing some sort of git action.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,6 +1,7 @@
 import cp = require('child_process')
 import fs = require('fs')
 import p = require('path')
+import os = require('os')
 
 // Logger
 const l = (msg: string): void => console.log(`husky - ${msg}`)
@@ -78,6 +79,19 @@ ${cmd}
   )
 
   l(`created ${file}`)
+
+  if (os.type() === 'Windows_NT') {
+    l(
+      `Due to a limitation on Windows systems, the executable bit of the file cannot be set without using git. 
+      To fix this, the file ${file} has been automatically moved to the staging environment and the executable bit has been set using git. 
+      Note that, if you remove the file from the staging environment, the executable bit will be removed. 
+      You can add the file back to the staging environment and include the executable bit using the command 'git update-index -add --chmod=+x ${file}'. 
+      If you have already committed the file, you can add the executable bit using 'git update-index --chmod=+x ${file}'. 
+      You will have to commit the file to have git keep track of the executable bit.`,
+    )
+
+    git(['update-index', '--add', '--chmod=+x', file])
+  }
 }
 
 export function add(file: string, cmd: string): void {


src/index.ts
```diff
@@ -78,6 +79,19 @@ ${cmd}
...
   l(`created ${file}`)
+
+  if (os.type() === 'Windows_NT') {
+    l(
+      `Due to a limitation on Windows systems, the executable bit of the file cannot be set without using git. 
+      To fix this, the file ${file} has been automatically moved to the staging environment and the executable bit has been set using git. 
+      Note that, if you remove the file from the staging environment, the executable bit will be removed. 
+      You can add the file back to the staging environment and include the executable bit using the command 'git update-index -add --chmod=+x ${file}'. 
+      If you have already committed the file, you can add the executable bit using 'git update-index --chmod=+x ${file}'. 
+      You will have to commit the file to have git keep track of the executable bit.`,
```
nathanpovo, line 90: I added a long log message here explaining the reasoning of why the new commit hook file is suddenly in the git staging environment. Feel free to modify this or give any suggestions on how this can be improved.

typicode:
Thank you! Looks great. I may revisit it a bit. The feature will be merged for the next major release (planned for Node 14 end of life or earlier).

SinaKarimi7:
It has been a problem for me until now. With a project created in Windows and configured with Husky and Commitlint, I noticed another team member could commit to remote without any restrictions (no Commitlint rules). While looking for the cause, I found that the files 'pre-commit' and 'commit-msg' aren't executable in Linux. Additionally, the `#!/usr/bin/env sh` must be changed to `#!/usr/bin/sh` because their distro (Manjaro I think) couldn't recognize and execute the file.

After changing the Shebang and exec permission the husky flow works.

yuntian001:
I need it

znzn007007:
great feature

TerryChenUI:
has it published?

PR #1278, opened 2023-07-03
docs: add more details

louis-vinchon:
Add more details on the purpose or added value the pacakge brings compared to plain git hooks.

To someone like me that has never used git hooks within a team, I scratched my head for a good 15 minutes trying to figure out what Husky was doing that I couldn't already do with plain git hooks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,9 @@
 
 > Modern native Git hooks made easy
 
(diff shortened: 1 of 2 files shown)


rickvian:
what about plain simple example of how to use it?


"husky": {
    "hooks": {
      "pre-commit": "your-command-here"
    }
  }

PR #1283, opened 2023-07-14
chore: added peer dep for eslint + Integration into CI

mastrzyz:
Added 'eslint' as a devdependency for package.json to eliminate future confusion.

Added it also as part of regular CI Workflows in github.


typicode:
Thank you for the PR, sorry I didn't have the time to review it before.

PR #1289, opened 2023-07-28
Update README.md

NomanGhiasy:
Added a step for enabling Git hooks


typicode:
Thank you for the PR. For now, I prefer to keep the doc as it is :+1:

PR #1293, opened 2023-08-07
Fix text (negation)

alex-gru:
can't vs can


typicode:
No error, it's can :) (command is not found in the Git app but is found in the terminal)

PR #1304, opened 2023-09-22
fix: Do not search for rc file at `/husky/init.sh`

hyperupcall:
Under most systems, `$XDG_CONFIG_HOME` is not defined. Therefore, in the most common case, Husky would check for the following files in the order:

- `/husky/init.sh`
- `$HOME/.config/husky/init.sh`
- `$HOME/.huskyrc.sh`

Of course, this isn't compliant with the XDG Directory Specification. This fixes the case whwere `XDG_CONFIG_HOME` is not defined - it defaults it to `$HOME/.config`, as per the specification.

### DIFF
--- a/husky.sh
+++ b/husky.sh
@@ -14,7 +14,7 @@ if [ -z "$husky_skip_init" ]; then
     exit 0
   fi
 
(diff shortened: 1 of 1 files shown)


typicode:
Thank you!

PR #1307, opened 2023-10-12
docs: prefer delete script prepare compare to ignore scripts

emmanuelgautier:
`--ignore-scripts` disable dependencies script as well. This argument prevent some dependencies to download binary during installation.


typicode:
Thank you

PR #1310, opened 2023-10-25
docs: use NODE_ENV to conditionally install husky

forivall:
See npm's documentation at https://docs.npmjs.com/cli/v8/commands/npm-install#omit

> If the resulting omit list includes `'dev'`, then the `NODE_ENV` environment
> variable will be set to `'production'` for all lifecycle scripts.


typicode:
Thank you!

PR #1325, opened 2023-12-12
docs: add favicon

rakleed:


typicode:
Thank you. I've published a new version, could you please add the change to `config.mts` instead (feel free to open a new PR) and use inline image, like this:

```
	head: [
		['link', { rel: 'icon', href: 'data:image/svg+xml,<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 100 100"><text y=".9em" font-size="85">🐶</text></svg>' }],
	],
```

PR #1333, opened 2024-01-25
v9

typicode:


typicode:
LGTM :rocket:

bin.js
```diff
@@ -0,0 +1,21 @@
...
+  process.stdout.write(i())
+  w('.husky/pre-commit', process.env.npm_config_user_agent.split('/')[0] + ' test')
+  process.exit()
+}
+
+let d = c => console.error(`${c} command is deprecated`)
+if (['add', 'set', 'uninstall'].includes(a)) { d(a); process.exit(1) }
+if (a == 'install') d(a)
+
+process.stdout.write(i(a == 'install' ? undefined : a))
```
Den-dp, line 21: I am curious, what guided you to go so far with reducing the package size? Isn't keeping the code readable more valuable than reducing the already small size?

Thanks for your work tho!

typicode: Good question and the answer is it depends.

Personally, I did it for fun (it's important sometimes) and for some people, size is an important metric.

> Isn't keeping the code readable more valuable than reducing the already small size?

I agree, that's why I'm not doing this for any other projects that I maintain. But here, husky is quite simple so I feel it's quite manageable.

Also, I found out some positive things:
- maintenance got a lot easier without TS, Prettier and ESLint dependencies. I like this simplicity.
- I don't have to build the project to publish it
- it forces me to focus on core value and avoid feature creep
- I feel that it's easier to understand at a glance what each line does without multi-lines if/else (but I'm biased since I wrote it)... also all the code fits on one screen, no scrolling necessary

Once again, I wouldn't do it for another project but here I'm satisfied with the end result.

> Thanks for your work tho!

Thanks for the kind words :)

Issue #1334, opened 2024-01-25
husky init not working


sinbino:
I am using docker node:20-slim image (debian 12.4, node 20.11.0, pnpm 8.14.3).
I get an error when I follow the [get started](https://typicode.github.io/husky/get-started.html) procedure.
The version of husky is 9.0.1

```
pnpm add --save-dev husky
pnpm exec husky init

node:internal/fs/utils:699
    throw new ERR_INVALID_ARG_TYPE(propName, ['string', 'Buffer', 'URL'], path);
          ^

TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string or an instance of Buffer or URL. Received undefined
    at Object.openSync (node:fs:579:10)
    at Object.readFileSync (node:fs:457:35)
    at file:///home/node/app/node_modules/.pnpm/husky@9.0.1/node_modules/husky/bin.js:9:24
    at ModuleJob.run (node:internal/modules/esm/module_job:218:25)
    at async ModuleLoader.import (node:internal/modules/esm/loader:329:24)
    at async loadESM (node:internal/process/esm_loader:28:7)
    at async handleMainPromise (node:internal/modules/run_main:113:12) {
  code: 'ERR_INVALID_ARG_TYPE'
}
```

recallwei:
Same to me, It seems `pnpm` not working, I use `npx` instead.

marcinincreo:
`husky.sh` file that init creates is empty when using `npx`

rozbo:
+1

dmitriy-kudelko:
Same here

artursudnik:
+1

denisrudov:
just fixed in 9.0.2 👍🏻

typicode:
It should be good now

PR #1347, opened 2024-01-25
fix: fix init not work on pnpm #1334

rozbo:
`npm_package_json`  env is not supported by pnpm now. (but npm and yarn works)

This is not a perfect fix, the user may not run husky init in project root. in this  case, we need more code that npm and yarn user doest't need, I think this is unfair.  So, we can simple add one line here utill the pnpm support it.

The other way is update doc, let users define a env named `npm_package_json` before `husky init`, but I think that is too complex for users.

### DIFF
--- a/bin.js
+++ b/bin.js
@@ -5,7 +5,7 @@ import i from './index.js'
 let a = process.argv[2]
 
 if (a == 'init') {
-  let p = process.env.npm_package_json
+  let p = process.env.npm_package_json || './package.json'
   let d = JSON.parse(f.readFileSync(p))
   d.scripts ||= {}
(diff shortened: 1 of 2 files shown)


rozbo:
fullly agree, clean and beautiful now

bin.js
```diff
@@ -5,7 +5,7 @@ import i from './index.js'
 let a = process.argv[2]
 
 if (a == 'init') {
-  let p = process.env.npm_package_json
+  let p = 'package.json'
```
glensc, line 8: why not still use the env,  but fallback if missing:

```js
let p = process.env.npm_package_json || 'package.json'
```

PR #1348, opened 2024-01-25
fix: init create dir before

typicode:


bin.js
```diff
@@ -11,6 +11,7 @@ if (a == 'init') {
   d.scripts.prepare = 'husky'
   w('package.json', JSON.stringify(d, null, /\t/.test() ? '\t' : 2) + '\n')
   process.stdout.write(i())
+	try { f.mkdirSync('.husky') } catch {}
```
glensc, line 14: unidented commit. seems to use tab while rest of the code use spaces

typicode: That was a mistake. It was fixed by another commit (using tabs with .editorconfig)

glensc: Time to dogfood and setup git hooks to husky repo to auto lint? ;)

typicode: Ahah probably :D Dog fooding json-server already.

Issue #1351, opened 2024-01-25
Init Broken in 9.0.4


gflohr:
**Troubleshoot**
- [x] Before creating an issue, please check:
https://typicode.github.io/husky/troubleshoot.html

If you're migrating from husky 4, see:
https://typicode.github.io/husky/migrate-from-v4.html

**Context**

```$ npx husky 
init
file:///path/to/node_modules/husky/bin.js:10
  (d.scripts ||= {}).prepare = 'husky'
   ^

ReferenceError: Cannot access 'd' before initialization
    at file:///path/to/node_modules/husky/bin.js:10:4
    at ModuleJob.run (node:internal/modules/esm/module_job:217:25)
    at async ModuleLoader.import (node:internal/modules/esm/loader:316:24)
    at async loadESM (node:internal/process/esm_loader:34:7)
    at async handleMainPromise (node:internal/modules/run_main:66:12)

Node.js v20.8.1
```

Downgrading to 9.0.3 fixed the problem.

Felipellira:
Same here

sankaSanjeeva:
+1

typicode:
Sorry, it's fixed. I've added a test to avoid regressions.

sankaSanjeeva:
@typicode Thank you for your prompt response

PR #1352, opened 2024-01-25
fix(init): seperate out prepare script declarations and assignment #1351

stagefright5:


Pietrucci-Blacher:
#1353

typicode:
Thanks @stagefright5, I just forgot that a `;` was needed.

PR #1357, opened 2024-01-25
docs: update install script to mjs and fix CI checks

### LINKED ISSUE #1356: Cannot run CI compatible install script
The installation scripts compatible with CI need to be modified for these reasons.

### Must indicate ES Modules because it uses top-level await

``` bash
PS C:path\to\directory> npm i

> directory@1.0.0 prepare
> node .husky/install.js

C:path\to\directory\.husky\install.js:5
const husky = await import('husky')
              ^^^^^

SyntaxError: await is only valid in async functions and the top level bodies of modules
    at internalCompileFunction (node:internal/vm:77:18)
    at Module._compile (node:internal/modules/cjs/loader:1340:27)
    at Module._extensions..js (node:internal/modules/cjs/loader:1435:10)
    at Module.load (node:internal/modules/cjs/loader:1207:32)
    at Module._load (node:internal/modules/cjs/loader:1023:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:135:12)
    at node:internal/main/run_main_module:28:49

Node.js v20.11.0
npm ERR! code 1
npm ERR! path C:\path\to\directory
npm ERR! command failed
npm ERR! command C:\WINDOWS\system32\cmd.exe /d /s /c node .husky/install.js
```

### `husky()` is not a function, so it must use `husky.default()`

``` bash
PS C:path\to\directory> npm i

> directory@1.0.0 prepare
> node .husky/install.mjs
file:///C:/path/to/directory/.husky/install.mjs:6
husky()
^

TypeError: husky is not a function
    at file:///C:/path/to/directory/.husky/install.mjs:6:1

Node.js v20.11.0
npm ERR! code 1
npm ERR! path C:path\to\directory
npm ERR! command failed

```

### CI checks should be fixed
According to ci-info, when we use `process.env.CI`, we should use `!!(process.env.CI !== 'false' && process.env.CI)`.

https://github.com/watson/ci-info/blob/v4.0.0/index.js#L56-L69

starnayuta:
close #1356

### DIFF
--- a/docs/how-to.md
+++ b/docs/how-to.md
@@ -66,21 +66,21 @@ Modify the `prepare` script to never fail:
 "prepare": "husky || true"
 ```
 
-You'll still get a `command not found` error message in your output which may be confusing. To make it silent, create `.husky/install.js`:
+You'll still get a `command not found` error message in your output which may be confusing. To make it silent, create `.husky/install.mjs`:
 
 ```js
 // Skip Husky install in production and CI
-if (process.env.NODE_ENV === 'production' || process.env.CI === '1') {
-  process.exit(0)
+if (process.env.NODE_ENV === 'production' || !!(process.env.CI !== 'false' && process.env.CI)) {
+    process.exit(0)
 }
 const husky = await import('husky')
-husky()
+husky.default()
 ```
 
 Then, use it in `prepare`:
 
 ```json
-"prepare": "node .husky/install.js"
+"prepare": "node .husky/install.mjs"
 ```
 
 ## Testing Hooks Without Committing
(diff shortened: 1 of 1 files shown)


typicode:
Thank you for the PR. You're right about the CI check. However I prefer easy-to-understand code, and it should be compatible with most CIs, especially GitHub Actions.

Using `.mjs` is a better choice :+1: . I've simplified the code too. There's actually no need for `await`.

starnayuta:
Okay, I see what you're thinking.

How about changing `process.env.CI === '1'` to `process.env.CI === 'true'`?

GitHub Actions is also set to true.  
https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables

typicode:
Good catch, I thought for GitHub Actions it was 1.

starnayuta:
If you run import without husky installed, `Cannot find package ‘husky’` error will occur.

https://github.com/typicode/husky/blob/main/docs/how-to.md?plain=1#L73

PR #1360, opened 2024-01-26
fix: add missing "types" and "main" fields to package.json to allow TS default ex…

jerome-benoit:
…port usage

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
 		"husky": "bin.js"
 	},
 	"exports": "./index.js",
+	"types": "./index.d.ts",
 	"engines": {
 		"node": ">=18"
 	}
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
 	"bin": {
 		"husky": "bin.js"
 	},
+	"main": "./index.js",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -19,7 +19,8 @@
 		"husky": "bin.js"
 	},
```
jerome-benoit, line 20:
```suggestion
	},
	"main": "./index.js",
```

typicode:
Thank you for the PR :)

I tried using quickly husky as a module in a TS file and didn't get an error. `index.d.ts` seems to be the default https://www.typescriptlang.org/docs/handbook/declaration-files/dts-from-js.html#editing-the-packagejson

So except if I'm missing something, I don't think it's necessary to be explicit about it.

And for `main`, it's for CommonJS which is not supported by the project.

jerome-benoit:
The current way the default husky export is currently is not ESM compatible: 

https://github.com/SAP/e-mobility-charging-stations-simulator/blob/main/prepare.js

The expected implementation should avoid to use to .default() on imported ESM module. I will open an issue with the repro case.

PR #1369, opened 2024-01-28
Update index.md

khaledYS:
wrong url routing to wrong md file, changed that


khaledYS:
a problem occur due to wrong redirected url, I changed /get-started to become /get-started.md, so it can function better and can redirect to the right path

typicode:
Thank you

PR #1371, opened 2024-01-29
Clarify Yarn not supporting prepare

AlecRust:
It seems only Yarn v2 (and above) don't support `prepare`. Yarn Classic i.e. v1, does.

As such it seems "manual setup if using Yarn" should only be recommended if _not_ using Yarn Classic.

In my testing with Yarn Classic, `yarn husky init` works fine, as does the `prepare` step in `package.json`.

### DIFF
--- a/docs/how-to.md
+++ b/docs/how-to.md
@@ -214,7 +214,7 @@ Run the `husky` command once in your repo. Ideally, include it in the `prepare` 
 ```json [yarn]
 {
   "scripts": {
-    // Yarn doesn't support prepare script
+    // Yarn (v2) doesn't support prepare script
     "postinstall": "husky",
     // Include this if publishing to npmjs.com
     "prepack": "pinst --disable",


st0w:
To add to this discussion regarding Yarn - the documentation contradicts itself and is thus unhelpful.

On the manual setup (https://typicode.github.io/husky/how-to.html#manual-setup) with yarn selected, the first part suggests things to add to `package.json` and the first comment says `Yawn doesn't support prepare script`. The very next step is `Run prepare once: yarn run prepare`.

How does that make sense? What is the proper way to setup Husky with Yarn?

typicode:
it's a mistake `yarn`  is enough to trigger the `postinstall` script there:
https://typicode.github.io/husky/how-to.html#manual-setup

chrisEff:
@typicode I think you're missing the point here.
Husky is still using `prepare` (in auto install and it also recommends it in manual install), while Yarn versions >+ 2.0 don't support `prepare` anymore. Husky should use `postInstall` instead.

elvinristi:
> Husky is still using `prepare` (in auto install and it also recommends it in manual install), while Yarn versions >+ 2.0 don't support `prepare` anymore. Husky should use `postinstall` instead.

Using `postinstall` in module which is published as library, has an issue with that. If you install the library in the project, then it will also trigger `postinstall` of that, so you might get exit code 127 during the `yarn install`

Trying now to use `[ "${HUSKY:-1}" = 0 ] && exit 0; command husky 2>/dev/null && husky` (on `*nix` based )

PR #1376, opened 2024-02-01
Fix tab detection on install command

glensc:
See https://github.com/typicode/husky/commit/ec13855c167bcead06523ae43ca1b5152d77810f#r138138417


glensc:
@typicode not tested, but change should be fine

PR #1378, opened 2024-02-01
refactor: rename files

typicode:


mrkjdy:
I believe this broke imports of this package. Now I'm getting this TypeScript error:
image

`index.d.ts` needs to be renamed to `index.d.mts` as well

typicode:
Thanks do you want to make a PR?

glensc:
care to explain why the rename is necessary? what does it do (improve?, fix?)

typicode:
> care to explain why the rename is necessary? what does it do (improve?, fix?)

It simplifies package.json. For a tiny project like this, I felt it would be better.

It's not necessary by any means but mostly a preference.

PR #1382, opened 2024-02-04
chore: add node badge

btea:


YasharF:
Is this still needed?

btea:
Personally, I think it can be added.

PR #1402, opened 2024-03-06
Changed default hook script to avoid fail on Windows installations

Jxlle:
Recently in Husky v9, an issue came up on Windows installations using Husky (in my case, wsl on Windows), where the default script used by the hooks fails on some Git GUI's (in my case it was GitKraken using a Windows installed Git executable) because the path variable used in the script has a slash in the other direction (`\`). This commit changes this base script to the following:

```
#!/usr/bin/env sh
SUB=${0%/*}
. "${SUB%\\*}/h"
```
instead of:
```
#!/usr/bin/env sh
. "${0%/*}/h"
```

This should work with both types of path separators.

**For me, this change in script fixed all my issues but I'm not sure why this problem is even occurring. I saw that some of the hooks (such as pre-commit) DO use a slash in the other direction (`/`) within GitKraken. Using the hooks via the Git cli on my Windows machine does NOT give problems for the original script.**

_**Note**: Paths that contain both slashes might still have problems as the above script can delete two slashes instead of one (if they are different). To avoid this, an if statement can be added but I'm not sure if this is required due to not knowing how common this situation is._

### DIFF
--- a/index.mjs
+++ b/index.mjs
@@ -17,7 +17,7 @@ export default (d = '.husky') => {
 	f.mkdirSync(_(), { recursive: true })
 	w(_('.gitignore'), '*')
 	f.copyFileSync(new URL('husky', import.meta.url), _('h'))
-	l.forEach(h => w(_(h), `#!/usr/bin/env sh\nPATH="\${0%/*}"\n. "\${PATH%\\\\*}/h"`, { mode: 0o755 }))
+	l.forEach(h => w(_(h), `#!/usr/bin/env sh\nPATH=\${0%/*}\n. "\${PATH%\\\\*}/h"`, { mode: 0o755 }))
 	w(_('husky.sh'), '')
 	return ''
 }
--- a/index.mjs
+++ b/index.mjs
@@ -17,7 +17,7 @@ export default (d = '.husky') => {
 	f.mkdirSync(_(), { recursive: true })
 	w(_('.gitignore'), '*')
 	f.copyFileSync(new URL('husky', import.meta.url), _('h'))
-	l.forEach(h => w(_(h), `#!/usr/bin/env sh\n. "\${0%/*}/h"`, { mode: 0o755 }))
+	l.forEach(h => w(_(h), `#!/usr/bin/env sh\nPATH="\${0%/*}"\n. "\${PATH%\\\\*}/h"`, { mode: 0o755 }))
 	w(_('husky.sh'), '')
 	return ''
 }
--- a/index.mjs
+++ b/index.mjs
@@ -17,7 +17,7 @@ export default (d = '.husky') => {
 	f.mkdirSync(_(), { recursive: true })
 	w(_('.gitignore'), '*')
 	f.copyFileSync(new URL('husky', import.meta.url), _('h'))
-	l.forEach(h => w(_(h), `#!/usr/bin/env sh\nPATH=\${0%/*}\n. "\${PATH%\\\\*}/h"`, { mode: 0o755 }))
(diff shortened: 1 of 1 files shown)


SoyDiego:
> Recently in Husky v9, an issue came up on Windows installations using Husky (in my case, wsl on Windows), where the default script used by the hooks fails on some Git GUI's (in my case it was GitKraken using a Windows installed Git executable) because the path variable used in the script has a slash in the other direction (`\`). This commit changes this base script to the following:
> 
> ```
> #!/usr/bin/env sh
> SUB=${0%/*}
> . "${SUB%\\*}/h"
> ```
> 
> instead of:
> 
> ```
> #!/usr/bin/env sh
> . "${0%/*}/h"
> ```
> 
> This should work with both types of path separators.
> 
> **For me, this change in script fixed all my issues but I'm not sure why this problem is even occurring. I saw that some of the hooks (such as pre-commit) DO use a slash in the other direction (`/`) within GitKraken. Using the hooks via the Git cli on my Windows machine does NOT give problems for the original script.**
> 
> _**Note**: Paths that contain both slashes might still have problems as the above script can delete two slashes instead of one (if they are different). To avoid this, an if statement can be added but I'm not sure if this is required due to not knowing how common this situation is._

Hi, I was watching your PR and I don't know if I have the same problem.
My project is like the documentation:

image

- I installed `husky`
- I did `npx husky init`
- Last command, edited my `package.json` and I replaced the line `prepare` for `"prepare": "cd .. && husky 'My Project/.husky'"`

When I do this, only is creating a file `pre-commit` but I cannot see the folder `_`

image


I'm doing manually `npm run prepare` and now I can see the folder `_`

image

Finally I'm trying to do a commit but always I have the same error:

image


What I'm doing wrong. I'm trying in Linux, Mac and Windows and always I have the same problem or sometimes I'm receiving the message `.git not found`.

Any help? @Jxlle @typicode 

Thanks!

ccerrillo (changes requested):
Hi @Jxlle, thanks for this!

I think a better aproach is to replace
`"\${0%/*}/h"`
 with
 `"\$(dirname "\$0")/h"`
 
Can you change this?

@typicode What do you think about this? Can it be merged with this change?

Jxlle:
> I think a better aproach is to replace `"\${0%/*}/h"` with `"\$(dirname "\$0")/h"`
> 
> Can you change this?

Hi @ccerrillo, do you mean like this?:

`#!/usr/bin/env sh\nSUB=\$(dirname "\$0")\n. "\${SUB%\\\\*}/h"`

which is

```
#!/usr/bin/env sh
SUB=$(dirname "$0")
. "${SUB%\\*}/h"
```

ccerrillo:
Hi @Jxlle,

No, with dirname there is no need to apply the regex to replace last part. just replace 
`"\${0%/*}/h"`
with
`"\$(dirname "\$0")/h"`

So the result is:

```
#!/usr/bin/env sh
. "\$(dirname "\$0")/h"
```

ccerrillo:
Hi @Jxlle I have found a way to completely solve the issue. The problem arises because GitKraken calls the script with the Windows path format, so using cygpath when available fixes it, along with using dirname in Husky hooks.
This is the pull request #1458 
@typicode can you look at it?

Jxlle:
> Hi @Jxlle I have found a way to completely solve the issue. The problem arises because GitKraken calls the script with the Windows path format, so using cygpath when available fixes it, along with using dirname in Husky hooks. This is the pull request #1458 @typicode can you look at it?

Looks great! Guess I'll close this PR as soon as @typicode has taken a look at yours

Issue #1387, opened 2024-02-06
Husky replaces my "prepare" script


ezeamin:
**Context**
- Terminal: Powershell - Windows 11
- Node: 20.9.0
- pnpm: 8.14.0

Hi! Husky is working great so far, thanks for your time! But, when I run `pnpm exec husky init`, following the docs, it installs the hooks but also replaces my `prepare` script with `prepare: husky`. I understand the purpose of this behaviour, but if I have a `prepare` script already, it should't replace it with its own script.

It also replaces my repo pre-commit file inside of `/husky` with a default one that does `pnpm test`. This is undesired behavior, because I already had a configuration written for the pre-commit hook.

typicode:
Hi!

For existing hooks, it's better to follow the 'How to migrate' section from https://github.com/typicode/husky/releases/tag/v9.0.1

I'll update docs to make that clearer.

deleonio:
Hello @typicode and @ezeamin,

we use the script `prepare` for more than just `husky`. I think it is more comfortable to leave it as it is.

zayne-wang:
@typicode
I created a pull request #1407  to fix this issue, please test and check.

SH4LIN:
At least there should be a condition to check if prepare script has any commands in it or not If there is a command then this `husky` command should be appended.

Like this

```Diff
- 14 ;(o.scripts ||= {}).prepare = 'husky'
+ 14 ;(o.scripts ||= {}).prepare = '' === (o.scripts ||= {}).prepare ? 'husky' : (o.scripts ||= {}).prepare + ' && husky'
```

zayne-wang:
> At least there should be a condition to check if prepare script has any commands in it or not If there is a command then this `husky` command should be appended.
> 
> Like this
> 
> ```diff
> - 14 ;(o.scripts ||= {}).prepare = 'husky'
> + 14 ;(o.scripts ||= {}).prepare = '' === (o.scripts ||= {}).prepare ? 'husky' : (o.scripts ||= {}).prepare + ' && husky'
> ```
> 
> Raised the PR for the mentioned solution please check it out.

Hi @SH4LIN,

Maybe we experienced the same mistake. According to typicode's reply in PR #1407, the custom scripts should be placed in `.husky/`, but not in `.husky/_/`. The `.husky/_/` directory is managed by husky it self.

Hope this could make sense for you.

benwck:
[...]

PR #1407, opened 2024-03-12
fix: husky will replace existing hook #1387 

zayne-wang:
fix: fix the issue #1387 
- add file detection, when file exists, the new file will not be created.

### DIFF
--- a/docs/package-lock.json
+++ b/docs/package-lock.json
@@ -2161,15 +2161,6 @@
         "node": ">=6"
       }
     },
-    "node_modules/escalade": {
-      "version": "3.1.1",
-      "resolved": "https://registry.npmjs.org/escalade/-/escalade-3.1.1.tgz",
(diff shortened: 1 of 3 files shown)


typicode:
Thank you for the PR. Actually files in `.husky/_/` are owned by husky and shouldn't be edited. That's why they get replaced.

User hooks needs to be in `.husky/`. For example, if you create `.husky/pre-commit` it won't be replaced.

rdmark:
@typicode Thanks for an awesome tool!

> User hooks needs to be in .husky/. For example, if you create .husky/pre-commit it won't be replaced.

This does not match what I'm seeing with husky 9.0.11. I have my own `.husky/pre-commit` file with custom commands. When I run `npx husky init` this file gets overwritten with the default `npm test` again.

I wonder if this is a bug, or some kind of misunderstanding?

typicode:
Some kind of misunderstanding. `husky init` is an automatic setup and should be run once (like `npm init` or `git init` for example).

PR #1458, opened 2024-06-12
Get the path from dirname instead of using Bash variable operators

ccerrillo:
On Windows, under certain circumstances, the path separator is "\".

This change will ensure compatibility with both Linux and Windows path separators


typicode:
Thank you!
Not sure how to test the cygwin code or if I want to support it for now considering the alternatives

PR #1479, opened 2024-07-26
Update pre-commit hook to run package commands directly

k-taro56:


typicode:
Thanks for the PR. Unfortunately it won't work.
It works with commands that are in `node_modules/.bin` and there's not `test` command there.

PR #1484, opened 2024-07-31
docs: update troubleshoot according not executable scripts

collierrgbsitisfise:


typicode:
Except if a bug, there's no need for them to be executable anymore

collierrgbsitisfise:
> Except if a bug, there's no need for them to be executable anymore. Edit: need to check the other PR actually...

@typicode why exactly "there's no need for them to be executable anymore"? When it was changed, probably i missed something 😅. Thx in advance!

typicode:
Sure, this code `sh -e "$s" "$@"` runs the hook script even if it's not executable.

Previously (before `v9`) it's true that, for example, `.husky/pre-commit` had to be executable. It was creating problems for people creating hooks on Windows due to missing executable permission.

collierrgbsitisfise:
closing this PR, cause it's not actual for latest version

PR #1497, opened 2024-08-20
fixes #1494, support pre-merge-commit hook

### LINKED ISSUE #1494: pre-merge-commit hook not triggering
I've set up a `pre-merge-commit` hook using Husky, but it doesn't seem to be triggering when I perform a merge. Here are the details:

## Steps to Reproduce:

1. Add a pre-merge-commit hook with the following command:

```sh
echo "echo 'pre-merge-commit hook triggered'" > .husky/pre-merge-commit
```

1. Merge a branch that should trigger the pre-merge-commit hook.
2. Notice that the hook does not run.

## Expected Behavior:
The `pre-merge-commit` hook should trigger and display the message `pre-merge-commit hook triggered` before the merge is completed.

## Actual Behavior:
The `pre-merge-commit` hook is not triggered at all, and the merge process continues without any output from the hook.

## Environment:
Husky Version: ^9.1.4
Node Version: 20.8.0
Git Version: 2.46.0
Operating System: MacOS Sonoma 14.0

### Workaround
There is a workaround to fix this issue by manually adding a `pre-merge-commit` file in the `./husky/_` directory with the following content:
```sh
#!/usr/bin/env sh
. "$(dirname "$0")/h"
```
After ensuring the file is executable, the issue is resolved. However, when the `_` directory is deleted and Husky is re-initialized using the `npm run prepare` command, the pre-merge-commit file is not generated. This defeats the purpose of using Husky, as the hook needs to be manually recreated each time.

RainMeoCat:
support pre-merge-commit hook

fixes: #1494

### DIFF
--- a/index.js
+++ b/index.js
@@ -2,7 +2,7 @@ import c from 'child_process'
 import f, { readdir, writeFileSync as w } from 'fs'
 import p from 'path'
 
-let l = [ 'pre-commit', 'prepare-commit-msg', 'commit-msg', 'post-commit', 'applypatch-msg', 'pre-applypatch', 'post-applypatch', 'pre-rebase', 'post-rewrite', 'post-checkout', 'post-merge', 'pre-push', 'pre-auto-gc' ],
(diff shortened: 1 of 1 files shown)


typicode:
Thank you

PR #1503, opened 2024-09-03
Fix how-to.html example for CI

mattixittam:
In the CI example, the docs incorrectly assume that the default import from the husky package is the install() method, where it is actually an object that has install() as a property. This cost me about 10 minutes just now, so I thought I'd fix the text.


typicode:
Thanks for the PR. Maybe you were on an older version.
There's no `install` export https://github.com/typicode/husky/blob/main/index.js#L8

PR #1519, opened 2024-09-11
Fix issue where example pre-commit file is generated incorrectly

dexmlee:
Fix to `bin.js` which incorrectly creates a partial example `pre-commit` file when `npx husky init` is called on systems with npm_config_user_agent set.


typicode:
Thank you

Issue #1520, opened 2024-09-16
Subsequent Init runs would clear Husky pre-commit 


fgrande:
Hi,

if I run "npx run init" after I've already changed pre-commit file in .husky/pre-commit, I lose my changes, it return to the "npm test"

I would suggest (going to send a PR) to create the pre-commit file only if it doesn't exists (so accidental multiple "npx run init" don't break changes).

Going to move the file creation into the try...catch, adding the flag "wx" (it raise an error if the file already exists)

regards
FabioG

Munaf-Divan:
I have a similar requirement. I want my existing `pre-commit` to remain unchanged while I run `npx husky init`.

typicode:
Init should be run only once. You can do the setup manually also.

fgrande:
@typicode It "should" be run only once. But what if a dev runs it again, after changes on pre-commit file ? That's the meaning of my PR.

shdq:
Removing the lines below from your hook files will solve your problems. See "How to Migrate" in [the release notes](https://github.com/typicode/husky/releases/tag/v9.0.1).

e.g. `.husky/pre-commit`

```diff
- #!/usr/bin/env sh
- . "$(dirname -- "$0")/_/husky.sh"
npm run lint --fix
```

DevJake:
> Deprecating install and replacing it with something that does not do what install did is annoying.

Exactly this. An `install` command is logical, and I don't understand why it was deprecated. I cannot imagine many would want to manually install hooks either. Really, the `init` command should be idempotent by not overwriting file contents. @shdq I have just tested with what you suggested, with no luck... the file contents are still being overwritten :/

The documentation is rather lacking on what an alternative solution would be. There seemingly isn't a `help` command (I might be wrong!), so that would be beneficial to view what commands and flags are available.

PR #1521, opened 2024-09-16
Subsequent Init runs would clear Husky pre-commit #1520

fgrande:
Hi,
as stated in #1520 I'm sending You a PR.

In my case, we share the same project between many coleagues. If one would run - erroneously - the "init", would delete the pre-commit hook. This is why - IMHO - would be fine to preserve it.

Regards
FabioG

### DIFF
--- a/bin.js
+++ b/bin.js
@@ -11,11 +11,13 @@ if (a == 'init') {
 	n = 'package.json'
 	s = f.readFileSync(n)
 	o = JSON.parse(s)
-	;(o.scripts ||= {}).prepare = 'husky'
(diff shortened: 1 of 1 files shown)


Munaf-Divan:
This is a good use case and addresses my needs.

fgrande:
@Munaf-Divan  I haven't added a new test for this in my PR. Maybe it can be nice to have (let me see if I have time to implement it)

AspadaX:
Can anyone please merge this PR. This feature is really helpful. The current version almost renders our team project's pre-commit unusable.

ben-buitendijk:
I would also like to see this merged!

Issue #914, opened 2021-03-28
`husky install` fails when using `--prod`


ext:
I have a `package.json` similar to:

```js
{
  "scripts": {
    "prepare": "husky install"
  },
  "devDependencies": {
    "husky": "^5.2.0"
  }
}
```

Running `npm install --prod` (or `NODE_ENV=production`) will only install production dependencies and thus `node_modules/.bin/husky` will not be present. In my case this happens when trying to build a docker container where I want to install only the production dependencies.

    $ npm install --prod

    > foobar@1.0.0 prepare
    > husky install

    sh: line 1: husky: command not found
    npm ERR! code 127
    npm ERR! path /home/ext/foobar
    npm ERR! command failed
    npm ERR! command sh -c husky install

    npm ERR! A complete log of this run can be found in:
    npm ERR!     /home/ext/.npm/_logs/2021-03-28T15_18_18_181Z-debug.log

What is the suggested way to handle this scenario?

With husky 4 it worked properly as `husky install` wasn't needed.

pinalbhatt:
adding --ignore-scripts worked for me 
`
RUN npm ci --only=production --ignore-scripts
`

ext:
Thank you for the tips, maybe it could be added to the docs in case others face the same issue?

typicode:
Yes, good suggestion. If anyone wants to do it?

rdmurphy:
I had a similar issue and the problem with using a blanket `--ignore-scripts` is it affects **all** packages being installed. So for example, I had a project that needed scripts to run for installing a binary (like [imagemin-jpegtran](https://github.com/imagemin/imagemin-jpegtran)) because a package using it appeared in `dependencies`.

I don't know what the solution is, but it'll be difficult to consider `husky` as an option if using `--production` will only work if _no_ other dependencies **do** need scripts to run on install, which is often out of my hands.

robhowell:
I just hit this problem found this issue, and discovered that there is an official recommended approach, using the `is-ci` module: https://typicode.github.io/husky/#/?id=disable-husky-in-cidocker

ext:
> I just hit this problem found this issue, and discovered that there is an official recommended approach, using the `is-ci` module: https://typicode.github.io/husky/#/?id=disable-husky-in-cidocker

Most of the linked approaches wont help, `--ignore-scripts` is the best (only?) option so far.

[...]

PR #1539, opened 2024-11-19
fix: insert husky into prepare command ci proof

lauhon:
Matching: https://typicode.github.io/husky/how-to.html#ci-server-and-docker

And: https://github.com/typicode/husky/issues/914#issuecomment-2145063108


lauhon:
Hey @typicode, any chance you could have a look and maybe merge this?

YasharF:
I am not sure if this is a good idea because it can mask issues in all environments, not just CI.

PR #1588, opened 2025-05-24
fix: improve Node.js version compatibility for scripts setup

Liam-Weitzel:
## Problem
The current code uses the nullish coalescing assignment operator (`||=`) which is only supported in Node.js 15.0.0 and later versions. This causes syntax errors in projects using older Node.js versions.

## Solution
Replace the `||=` operator with standard JavaScript OR operator pattern that has wider compatibility:
```javascript
o.scripts = o.scripts || {};
o.scripts.prepare = 'husky';
```

## Testing
- None

### DIFF
--- a/bin.js
+++ b/bin.js
@@ -11,7 +11,8 @@ if (a == 'init') {
 	n = 'package.json'
 	s = f.readFileSync(n)
 	o = JSON.parse(s)
-	;(o.scripts ||= {}).prepare = 'husky'
+	o.scripts = o.scripts || {};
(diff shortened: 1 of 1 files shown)


YasharF:
Node 18+ is currently required: https://github.com/typicode/husky/blob/0692639896ea5b5dd5ffc607575f6f870b21cb0f/package.json#L23

The package is currently only being tested and released for such versions:
https://github.com/typicode/husky/blob/main/.github/workflows/node.js.yml

